PR #358, opened 2017-09-28
More SQ fixes

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/bean/ModuleInterface.java
```diff
@@ -68,10 +68,7 @@ public void setVersion(String version) {
    */
   public static boolean validateVersion(String version) {
     int[] p = versionParts(version, 0);
-    if (p == null) {
-      return false;
-    }
-    return true;
+    return p.length == 0 ? false : true;
   }
```
heikkilevanto: Old code: if p==null return false. New code refers to p.length, will that not crash on null ptr?

PR #438, opened 2017-11-10
FOLIO-462 perl one liner

dcrossleyau:
Extract relevant fenced code blocks. Added syntax highlighting marker “script”.


dcrossleyau:
That is OKAPI-462

PR #441, opened 2017-11-14
OKAPI-462 perl one-liner 2

dcrossleyau:
Tidy.


dcrossleyau:
That seems to be a Jenkins problem with running sonarqube.

PR #461, opened 2017-12-13
Allow content-length header

cherewaty:
We were having trouble receiving params in [`mod-kb-ebsco`](https://github.com/thefrontside/mod-kb-ebsco) when it's installed as part of an Okapi cluster.

@sadatay diagnosed the problem in `mod-kb-ebsco`'s Ruby stack: Rack uses `Content-Length` to determine whether or not it should attempt to parse the request body into params. Since `Content-Length` never makes it past Okapi, Rack and ActionDispatch behave as if they've seen a zero-length request body.

Passing through the `Content-Length` header to modules should fix the issue.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
@@ -432,6 +432,7 @@ public class MainVerticle extends AbstractVerticle {
             .allowedMethod(HttpMethod.POST)
             //allow request headers
             .allowedHeader(HttpHeaders.CONTENT_TYPE.toString())
+            .allowedHeader(HttpHeaders.CONTENT_LENGTH.toString())
             .allowedHeader(XOkapiHeaders.TENANT)
             .allowedHeader(XOkapiHeaders.TOKEN)
             .allowedHeader(XOkapiHeaders.AUTHORIZATION)


cherewaty:
It looks like `Content-Length` headers are also blacklisted in a couple places?
https://github.com/folio-org/okapi/blob/1d0068d4e13308e425a4e6ba24f2334b21e28f56/okapi-core/src/main/java/org/folio/okapi/service/ProxyService.java#L364
https://github.com/folio-org/okapi/blob/1d0068d4e13308e425a4e6ba24f2334b21e28f56/okapi-core/src/main/java/org/folio/okapi/service/ProxyService.java#L667

@adamdickmeiss why does Okapi need to remove `Content-Length`?

adamdickmeiss:
We used chunked encoding for Okapi . This is HTTP/1.1 and overrides content-length and so to avoid confusion we remove it to make it 100% clear that length is given by chunks. I don't even think removing the two lines above solves your problem. Have you determined that it DOES or does it just appear to do so - from code inspection?

adamdickmeiss:
We'll get back to you on this.

cherewaty:
@adamdickmeiss you're right, this PR will probably not solve our problem. `Transfer-Encoding: Chunked` was the missing piece in our debugging. Closing this PR while we look for another solution.

adamdickmeiss:
Tjanks.. If it turns out to be a big problem I'm sure we can make a new routing type where Okapi piles up the full posted body before making the header so that the Content-Length can be set. (The chunked encoding makes it possible to "post" large bodies that don't need to be in memory, so that's why from a proxy perspective it's an efficient way to implement things).

PR #539, opened 2018-02-23
Clean up auth header processing: Send them only to auth filter, sanit…

heikkilevanto:
…ize away


heikkilevanto:
I will make a new PR when I fix Okapi-538. I will address the SQ issues in that as well

PR #668, opened 2018-08-03
OKAPI-635: Return Auth filter error without relying on pre/post filter

hjiebsco:
a bit more change


hjiebsco:
Merge now to unblock API test failures

PR #707, opened 2018-10-29
Okapi 674 proxy rewrite path

wwelling:


wwelling:
We discussed that we did not want the notion of basePath for interfaces in which we could extend to scope during the proxy to a module. Therefore, I just have it rewrite to a simple path. Do we want to make this more complex? We would have to analyze the pathPattern and determine which parts should be rewritten.

wwelling:
I would like to specify a basePath at the interface provides level. Then each path pattern of the handlers is concatenated with the basePath and used as normal. Then additionally a handler could overwrite the basePath to perform a desired rewrite.

wwelling:
We could simply just concat the path after the rewrite path. I will change it to do that.

adamdickmeiss:
Yep. Keeping the path there is better.

wwelling:
Ok. Ready for another review.

adamdickmeiss:
Looks good

adamdickmeiss:
I think squash and merge is fine, no need to save intermediate steps (with a changed feature)

PR #714, opened 2018-11-07
Okapi 678 incorrect content type for mixed content

adamdickmeiss:


wwelling:
This works great. Thank you.

PR #839, opened 2019-09-26
 using ip "0.0.0.0" as -cluster-host default value in cluster mode

x19990416:
run okapi in cluster mode, we need a paramter -clust-host to tell hazelcast which network is using.
so if we using ip "0.0.0.0" as default value, it will be use the local ip address.
this will be more smarter, the user no need to care param -cluster-host.


x19990416:
is there something wrong with my pr？

x19990416:
using 0.0.0.0 as default of  -cluster-host. so when running okapi with no -cluster-host param, it will use the default ip instead of "127.0.0.0" in cluster mode.that user does not need to care the ip address in the same net.

x19990416:
is there something wrong with my pr？

PR #840, opened 2019-09-26
Sanitize response headers OKAPI-764

adamdickmeiss:
It happened always betwen modules, but just not for the last
response.


adamdickmeiss:
> Calling sanitizeAuthHeaders is a good thing.
> But that does not seem to remove the X-Okapi-Token header, only the special headers used for the communication with mod-auth-token. Better check if we can still leak the X-Okapi-Header.

Yes. This does not remove X-Okapi-Token.. I though that should be a separte issue OKAPI-763 and so a separate commit/PR

PR #861, opened 2019-11-29
Okapi 780 switch to log4j2

adamdickmeiss:


adamdickmeiss:
No need to review this yet.. Need to fix the SQ warnings first.

okapi-core/src/main/java/org/folio/okapi/MainDeploy.java
```diff
@@ -162,7 +160,7 @@ private boolean parseOptions(String[] args, Handler<AsyncResult<Vertx>> fut) {
   }
 
   private void printUsage() {
```
julianladisch: Can you add

    @SuppressWarnings("squid:S106")  // Suppress "Standard outputs should not be used directly"

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -89,13 +89,13 @@ private void loadModules(Handler<ExtendedAsyncResult<Void>> fut) {
...
             } else {
-              CompList<Void> futures = new CompList<>(INTERNAL);
+              CompList<Void> futures = new CompList<>(ErrorType.INTERNAL);
               for (ModuleDescriptor md : mres.result()) {
                 Promise<Void> promise = Promise.promise();
                 modules.add(md.getId(), md, promise::handle);
                 futures.add(promise);
               }
-              futures.all(fut);
+              futures.seq(fut);
```
julianladisch: This change should be in a separate pull request and a separate jira issue.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -125,20 +129,16 @@ public void insert(Tenant t, Handler<ExtendedAsyncResult<String>> fut) {
     String id = t.getId();
     tenants.get(id, gres -> {
       if (gres.succeeded()) {
-        fut.handle(new Failure<>(USER, messages.getMessage("10400", id)));
-      } else if (gres.getType() == NOT_FOUND) {
-        if (tenantStore == null) { // no db, just add it to shared mem
```
julianladisch: The tenantStore parameter of the constructor may be null.
Can you add a null check to the constructor to fail the constructor with a NullPointerException?

okapi-core/src/main/java/org/folio/okapi/service/impl/PostgresHandle.java
```diff
@@ -56,35 +53,22 @@ protected PostgresHandle(Vertx vertx, JsonObject conf) {
         Integer x = Integer.parseInt(val);
         pgconf.put("port", x);
       } catch (NumberFormatException e) {
-        logger.warn("Bad postgres_port value: " + val + ": " + e.getMessage());
+        logger.warn("Bad postgres_port value: {}: {}", val, e.getMessage(), e);
```
julianladisch: Can you remove `, e`? In this case the stacktrace is useless.

okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
```diff
@@ -129,7 +129,8 @@ private void start2(Handler<AsyncResult<Void>> startFuture) {
       future.complete();
     }, false, result -> {
       if (result.failed()) {
-        logger.debug("ProcessModuleHandle.start2() executeBlocking failed " + result.cause());
+        logger.debug("ProcessModuleHandle.start2() executeBlocking failed {}",
+          result.cause());
```
julianladisch: result.cause().getMessage()

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -1240,13 +1215,7 @@ private void loadTenants(Handler<ExtendedAsyncResult<Void>> fut) {
...
-        if (!keys.isEmpty()) {
-          logger.info("Not loading tenants, looks like someone already did");
-          fut.handle(new Success<>());
-        } else if (tenantStore == null) {
-          logger.info("No storage to load tenants from, so starting with empty");
-          fut.handle(new Success<>());
-        } else {
+        if (keys.isEmpty()) {
           loadTenants2(fut);
         }
```
julianladisch: The else-clause with a fut.handle(...) is missing.

Consider using guard-clauses and merging laodTenants2 into loadTenants:

    tenants.getKeys(gres -> {
      if (gres.failed()) {
        fut.handle(new Failure<>(gres.getType(), gres.cause()));
        return;
      }
      Collection<String> keys = gres.result();
      if (! keys.isEmpty()) {
        logger.info("Not loading tenants, looks like someone already did");
        fut.handle(new Success<>());
        return;
      }
      tenantStore.listTenants(lres -> {
        if (lres.failed()) {
          fut.handle(new Failure<>(lres.getType(), lres.cause()));
          return;
        }
        CompList<List<Void>> futures = new CompList<>(ErrorType.INTERNAL);
        for (Tenant t : lres.result()) {
          Promise<Void> f = Promise.promise();
          tenants.add(t.getId(), t, f::handle);
          futures.add(f);
        }
        futures.all(fut);
      }
    }

okapi-core/src/main/java/org/folio/okapi/MainCluster.java
```diff
@@ -10,12 +9,14 @@ private MainCluster() {
...
   public static void main(String[] args) {
+    System.setProperty("vertx.logger-delegate-factory-class-name",
+     "io.vertx.core.logging.Log4jLogDelegateFactory");
     Logger logger = OkapiLogger.get();
     MainDeploy d = new MainDeploy();
     d.init(args, res -> {
       if (res.failed()) {
-        logger.error(res.cause().getMessage(), res.cause());
-        exit(1);
+        logger.error(res.cause().getMessage());
```
julianladisch: Removing `res.cause()` swallows the stacktrace. Why?

okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
```diff
@@ -134,7 +134,9 @@ private void logHandler(Buffer b) {
       logSkip = logSkip - b.length();
     }
     if (logBuffer.length() > 0 && logBuffer.charAt(logBuffer.length() - 1) == '\n') {
-      logger.info(id + " " + logBuffer.substring(0, logBuffer.length() - 1));
+      if (logger.isInfoEnabled()) {
+        logger.info("{} {}", id, logBuffer.substring(0, logBuffer.length() - 1));
+      }
```
julianladisch: Consider using lambdas to avoid the if-clause: https://logging.apache.org/log4j/2.0/manual/api.html#Java_8_lambda_support_for_lazy_logging

    logger.info("{} {}", () -> id, () -> logBuffer.substring(0, logBuffer.length() - 1));

okapi-core/src/main/java/org/folio/okapi/service/impl/MongoUtil.java
```diff
@@ -80,9 +79,9 @@ public void insert(T md, String id, Handler<ExtendedAsyncResult<Void>> fut) {
       if (res.succeeded()) {
         fut.handle(new Success<>());
       } else {
-        logger.warn("MongoUtil.insert " + id + " failed : " + res.cause());
-        logger.warn("Document: " + document.encodePrettily());
-        fut.handle(new Failure<>(INTERNAL, res.cause()));
+        logger.warn("MongoUtil.insert {} failed: {}", id, res.cause().getMessage());
+        logger.warn("Document: {}", document.encodePrettily());
```
julianladisch: Consider using a lambda to skip the encoding when logging warnings is disabled.

okapi-core/src/main/java/org/folio/okapi/service/impl/PostgresHandle.java
julianladisch: Can you remove `, e`? Adding a stacktrace for this NumberFormatException doesn't make sense.

okapi-core/src/main/java/org/folio/okapi/service/impl/Storage.java
```diff
@@ -72,20 +73,31 @@ public void prepareDatabases(InitMode initModeP, Handler<ExtendedAsyncResult<Voi
...
+      tenantStore.init(reset, promise.future());
+      return promise.future();
+    }).compose(res -> {
+      if (moduleStore == null) {
+        return Future.succeededFuture();
+      }
+      Promise promise = Promise.promise();
+      moduleStore.init(reset, promise.future());
+      return promise.future();
+    }).setHandler(fut);
```
julianladisch: Can you use the more idiomatic syntax explained on https://vertx.io/docs/vertx-core/java/#_sequential_composition ?

    .compose(res -> Future.future(promise -> envStore       .init(reset, (Future) promise.future())))
    .compose(res -> Future.future(promise -> deploymentStore.init(reset, (Future) promise.future())))
    .compose(res -> Future.future(promise -> tenantStore    .init(reset, (Future) promise.future())))

julianladisch: Or create a util class that directly supports a sequence of compose with ExtendedAsyncResult?

okapi-core/src/main/java/org/folio/okapi/MainCluster.java
adamdickmeiss: No stacktrace.. It says, for example, io.vertx.core.impl.NoStackTraceThrowable: Unknown command: xx , so it seems useless to me

okapi-core/src/main/java/org/folio/okapi/service/impl/Storage.java
adamdickmeiss: That would be nice..

PR #862, opened 2019-12-02
JavaDoc and note on id property of module descriptor in guide

adamdickmeiss:


doc/guide.md
```diff
@@ -187,7 +187,15 @@ be precise), and reports all "provided" interfaces.
 * `ModuleDescriptor.json` has a list of all `routes` (HTTP paths and
 methods) that a given module handles, this gives Okapi necessary
 information to proxy traffic to the module (this is similar to a
-simplified RAML specification).
+simplified RAML specification). Modules are identified in the `id`
+property of the module descriptor. It consists of two components:
+product and [semantic version](https://semver.org/) and separated by a
+dash. Think of product as implementation. There should only be one
```
julianladisch: Dash or hyphen-minus?
https://en.wikipedia.org/wiki/Dash
https://en.wikipedia.org/wiki/Hyphen-minus

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
```diff
@@ -7,6 +7,14 @@
   private SemVer semVer;
   private final String id;
 
+  /**
+   * Construct Module ID from Module ID string or product only (moduleID without
+   * version).
+   * May throw IllegalArgumentException for invalid syntax for semantic version.
+   * @param s Module ID or product name. The module ID does not have restrictions
+   * on characters, but a digit following a dash marks the beginning of a semantic
```
julianladisch:
```suggestion
   * on characters, but a hyphen-minus followed by a digit marks the beginning of a semantic
```

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
```diff
@@ -79,6 +116,9 @@ public boolean equals(Object that) {
     return compareTo((ModuleId) that) == 0;
   }
 
+  /**
+   * @return hash-code
+   */
```
julianladisch: Please do not overwrite the inherited Javadoc (`@Override`) unless the new Javadoc is better.

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
```diff
@@ -88,6 +128,10 @@ public int hashCode() {
     return c;
   }
 
+  /**
+   * String representation of module ID
+   * @return string representation
+   */
```
julianladisch: Please do not overwrite the inherited Javadoc (`@Override`) unless the new Javadoc is better.

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
```diff
@@ -99,12 +143,24 @@ public String toString() {
...
   public static int compare(String i1, String i2) {
     ModuleId m1 = new ModuleId(i1);
     ModuleId m2 = new ModuleId(i2);
     return m1.compareTo(m2);
   }
 
+  /**
+   * Returns newest module out of a list of modules
+   * @param l list of module IDs
+   * @return newest module (null for empty list)
```
julianladisch: No, it never returns null.
It first filters all modules where the procuct equals to this.getProduct(). If there is none it returns this.getId() otherwise it returns the newest of them.

okapi-common/src/main/java/org/folio/okapi/common/SemVer.java
```diff
@@ -74,14 +95,27 @@ private int parseComp(String v, boolean digits, int offset, List<String> result)
...
+  /**
+   * Whether version has pre-release
+   * @return true if version has pre-release; false otherwise
+   */
   public boolean hasPreRelease() {
     return !preRelease.isEmpty();
   }
 
+  /**
+   * Whether this version has prefix of other
```
julianladisch: Can you explain in the Javadoc what a prefix is in the context of semantic versioning? I cannot find it on https://semver.org/spec/v1.0.0.html

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
adamdickmeiss: damn right

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
adamdickmeiss: oh nice

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
adamdickmeiss: oh nice

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
adamdickmeiss: I sort of regret this feature.. It's offerd by Okapi so you can supply a fitler.. Eg "foo" will match module ID with product "foo". And "foo-1" will match anything with product "foo" and major=1.

adamdickmeiss: There's no such notion in semver

adamdickmeiss:
The SemVer.compareComp may look innocent, but I can see that sorting is not transitive.. SO wondering if I should change it.. semver doesn't really say.. testMixedPrerelease illustrates that.. Eg v1 > v2 and v2 > v3 and v3 > v4 and v4 > v1 !!!

okapi-common/src/main/java/org/folio/okapi/common/SemVer.java
adamdickmeiss: See comments for previous review item.. I typed in the wrong box

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
```diff
@@ -99,12 +136,24 @@ public String toString() {
...
+   * {@link org.folio.okapi.common.ModuleId#compareTo(org.folio.okapi.common.ModuleId)
+   */
   public static int compare(String i1, String i2) {
     ModuleId m1 = new ModuleId(i1);
     ModuleId m2 = new ModuleId(i2);
     return m1.compareTo(m2);
   }
 
+  /**
+   * Returns newest module out of a list of modules including this
```
julianladisch:
```suggestion
   * Returns newest module out of a list of modules where the product equals this.getProduct(); returns this.getId() if none is found
```

PR #863, opened 2019-12-10
Further testing of ModuleId; getLatest update

adamdickmeiss:


okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
```diff
@@ -155,17 +158,17 @@ public static int compare(String i1, String i2) {
    * @return newest module (possibly this module)
    */
   public String getLatest(Collection<String> l) {
-    ModuleId bModule = null;
-    String bId = null;
+    ModuleId bModule = this;
+    String bId = id;
```
julianladisch: There is no need to maintain bId. Simply return bModule.getId() at the end.

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
```diff
@@ -155,17 +158,17 @@ public static int compare(String i1, String i2) {
    * @return newest module (possibly this module)
    */
   public String getLatest(Collection<String> l) {
-    ModuleId bModule = null;
-    String bId = null;
+    ModuleId bModule = this;
```
julianladisch: A more self-explanatory name could be `latestModule`.

okapi-common/src/test/java/org/folio/okapi/common/ModuleIdTest.java
```diff
@@ -84,5 +85,10 @@ public void test2() {
     assertTrue(module_1_10000.hasSemVer());
     assertFalse(module_1_10000.hasPreRelease());
     assertTrue(module_1_10000.hasNpmSnapshot());
+
+    ModuleId module = new ModuleId("module");
+    assertFalse(module.hasSemVer());
+    assertFalse(module.hasPreRelease());
+    assertFalse(module.hasNpmSnapshot());
```
julianladisch: This new test should be in a separate unit test method, for example `withoutSemVer()`.

okapi-common/src/test/java/org/folio/okapi/common/ModuleIdTest.java
```diff
@@ -30,6 +30,7 @@ public void test() {
     versionsL.add("other-1.1");
     versionsL.add("other-0.9");
     assertEquals("module-1.0", module_1.getLatest(versionsL));
+    assertEquals("module-1", module_1.getLatest(new LinkedList<>()));
```
julianladisch: The getLatest unit tests should go into one or two separate unit test methods.

julianladisch: Thanks, resolved.

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
julianladisch: Thanks, this has been resolved.

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
julianladisch: Thanks, this has been resolved.

okapi-common/src/test/java/org/folio/okapi/common/ModuleIdTest.java
julianladisch: Thanks, this has been resolved.

PR #866, opened 2019-12-30
Okapi 789 max conn iteration config

fluagen:
add max_conn_iteration to vertx's config for feature OKAPI-789


adamdickmeiss:
Thx for the patch.. I moved it to a new PR.. https://github.com/folio-org/okapi/pull/872

PR #869, opened 2020-01-04
HTTP Client cache (GET, HEAD methods) OKAPI-770

adamdickmeiss:


okapi-common/src/main/java/org/folio/okapi/common/HttpClientCached.java
```diff
@@ -0,0 +1,286 @@
...
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.TreeSet;
+import org.apache.logging.log4j.Logger;
+
+/**
+ * HTTP client with cache .. Can be used in most case as a replacement of
+ * Vert.x {@link io.vertx.core.http.HttpClient}
+ *
+ * @author adam
```
julianladisch: The list of authors that contributes code is tracked in the git commit history.
Do you want to duplicate that information in the source code? How do we ensure that it is kept up to date? Is "adam" unambiguous now and in future?

adamdickmeiss: Yes.. I'll remove it and the 8 in total Okapi. 67 in RMB by the way

julianladisch: Thanks, resolved. (I cannot resolve this `@author` conversion because I don't have write permissions for this repository.)

PR #883, opened 2020-02-18
Okapi 796 reduce decoding module descriptors

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -38,6 +38,7 @@
   private String mapName = "modules";
   private LockedTypedMap1<ModuleDescriptor> modules
     = new LockedTypedMap1<>(ModuleDescriptor.class);
+  private Map<String,ModuleDescriptor> enabledModules = new HashMap<>();
```
julianladisch: Is this a cache? If yes, can you rename it to `enabledModulesCache`?

julianladisch: Is the cache pruned when ModuleManager.update or ModuleManager.delete is called, and are there unit tests for update and delete?

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -685,7 +685,7 @@ private void proxyRequestResponse10(Iterator<ModuleInstance> it,
 
   private void copyHeaders(HttpClientRequest cReq, RoutingContext ctx, ModuleInstance mi) {
     int sz = 0;
-    int limit = 5; // all headers dumped
+    int limit = 2000; // all headers dumped
```
julianladisch: Can you also change

    if (sz > limit) {

to

    if (sz > limit && logger.isInfoEnabled()) {

and remove the unused StringBuilder from dumpHeaders?

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
adamdickmeiss: yep

adamdickmeiss: In reality not needed, but I'll just do it. MDs are immutable and probably MD PUT will be removed soon.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: Yep. Will do

PR #884, opened 2020-02-21
Optional dependencies OKAPI-509

adamdickmeiss:


adamdickmeiss:
I can possibly improve the documentation with an example ..

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
```diff
@@ -220,23 +235,180 @@ public void test4(TestContext context) {
...
+      context.assertEquals(2, tml.size());
+      context.assertEquals("moduleA-1.1.0", tml.get(0).getId());
+      context.assertEquals("moduleA-1.0.0", tml.get(0).getFrom());
+      context.assertEquals("enable", tml.get(0).getAction().name());
+      context.assertEquals("moduleD-1.1.0", tml.get(1).getId());
+      context.assertEquals(null, tml.get(1).getFrom());
+      context.assertEquals("enable", tml.get(1).getAction().name());
+      async.complete();
+    });
+  }
```
julianladisch: There are two relevant cases that should also been tested:
If mdA100, mdA110, mdD100 and mdD110 are available, and mdA100 and mdD100 are enabled.
First case: Enable mdA110 and check that both are upgraded.
Second case: Enable mdD110 and check that both are upgraded.

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
```diff
@@ -220,23 +235,180 @@ public void test4(TestContext context) {
     });
   }
 
+  // install optional with no provided ingerface enabled
+  @Test
+  public void testInstallOptional1(TestContext context) {
+    Async async = context.async();
```
julianladisch: In all these tests there is no need to use Async.
Replace "res -> { context.assertTrue(res.succeeded());" by "context.asyncAssertSuccess(res ->"
and "res -> { context.assertTrue(res.failed());" by "context.asyncAssertFailure(exception ->"

adamdickmeiss: oh yes.. Seen that before

adamdickmeiss: Can't do like that, cause AsyncResult != ExtendedAsyncResult. I know one could use a utility or something, but least the async is systematically used.

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
adamdickmeiss: First case: only mdA110 will be upgraded, because mdD100 can use the provided interface of mdA110.

Second case: both will be upgraded, because mdD110 will require mdA110.

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
julianladisch: That's true and technical dept. https://issues.folio.org/browse/OKAPI-810 "Replace ExtendedAsyncResult by AsyncResult where possible"

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
```diff
@@ -220,23 +235,247 @@ public void test4(TestContext context) {
...
+    tm.setId(mdA110.getId());
+    tml.add(tm);
+
+    DepResolution.installSimulate(modsAvailable, modsEnabled, tml, res -> {
+      context.assertTrue(res.succeeded());
+      logger.debug("tml result = " + Json.encodePrettily(tml));
+      context.assertEquals(1, tml.size());
+      context.assertEquals("moduleA-1.1.0", tml.get(0).getId());
+      context.assertEquals("moduleA-1.0.0", tml.get(0).getFrom());
+      context.assertEquals("enable", tml.get(0).getAction().name());
```
julianladisch: Yes, no moduleD upgrade is needed because moduleA's new interface is compatible (int-1.0 -> int-1.1).

Can you add another test where moduleA-1.0.0 is upgraded to moduleA-2.0.0 with int-2.0 and this triggers the moduleD upgrade from moduleD-1.0.0 to moduleD-2.0.0 because moduleD-2.0.0 has optional dependency on int-2.0?

adamdickmeiss: Good one. That was a reverse dependency not catched.. Now fixed.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -458,14 +487,57 @@ private static void addOrReplace(List<TenantModuleDescriptor> tml, ModuleDescrip
...
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+      if (mTo != null) {
+        List<String> ret = addModuleDependencies(mTo, modsAvailable, modsEnabled, tml);
```
julianladisch: `ret` is not used. Can you remove this variable?

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -458,14 +487,57 @@ private static void addOrReplace(List<TenantModuleDescriptor> tml, ModuleDescrip
...
+                  if (prov.isCompatible(re1)) {
+                    if (mTo == null || ma.compareTo(mTo) > 0) {
+                      mTo = ma;
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
```
julianladisch: This will reduce the cyclomatic complexity from 52 to 37 -- less nesting improves readability:

      for (InterfaceDescriptor prov : md.getProvidesList()) {
        for (InterfaceDescriptor req : me.getRequiresOptionalList()) {
          if (! prov.getId().equals(req.getId()) || prov.isCompatible(req)) {
            continue;
          }
          for (ModuleDescriptor ma : modsAvailable.values()) {
            if (! me.getProduct().equals(ma.getProduct())) {
              continue;
            }
            for (InterfaceDescriptor re1 : ma.getRequiresOptionalList()) {
              if (prov.isCompatible(re1) && (mTo == null || ma.compareTo(mTo) > 0)) {
                mTo = ma;
              }
            }
          }
        }
      }

adamdickmeiss: Done

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: Done

PR #889, opened 2020-03-10
Okapi 773 switch discovery only if all succeeds

adamdickmeiss:
This PR improves coverage in general for install. It ensures that the tenant module association is not happening until all modules have been completing tenant init and tenant permissions interface. Of course, if the modules do not implement these interfaces, they will not be considered. Also if a module is not running at all, of course, also that stops upgrading. There's no switch for this. This changes behavior for the better, so a desired breaking change = bug fix.

Out of scope of this issue:

We should extend the tenant interface with a rollback mechanism so that module can be notified to rollback if they have succeeded but another module has failed. 

Finally, we could do a pre-check tenant init which has no side effects but just let the module ack that it can do upgrade (without actually doing anything). https://issues.folio.org/browse/OKAPI-814

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
@@ -1,6 +1,9 @@
 package org.folio.okapi;
 
-import org.folio.okapi.managers.ModuleManager;
+import java.lang.management.ManagementFactory;
+import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import io.vertx.core.AbstractVerticle;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
@@ -14,11 +17,6 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.core.spi.cluster.ClusterManager;
 import io.vertx.ext.web.Router;
 import io.vertx.ext.web.handler.CorsHandler;
-import static java.lang.System.getenv;
-import java.lang.management.ManagementFactory;
-import java.util.Set;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
 import org.apache.logging.log4j.Logger;
 import org.folio.okapi.bean.ModuleDescriptor;
 import org.folio.okapi.bean.Tenant;
@@ -28,9 +26,10 @@ import org.folio.okapi.common.Messages;
 import org.folio.okapi.common.ModuleVersionReporter;
 import org.folio.okapi.common.OkapiLogger;
 import org.folio.okapi.managers.DeploymentManager;
-import org.folio.okapi.service.ModuleStore;
+import org.folio.okapi.managers.ModuleManager;
 import org.folio.okapi.managers.ProxyService;
 import org.folio.okapi.managers.TenantManager;
+import org.folio.okapi.service.ModuleStore;
 import org.folio.okapi.service.TenantStore;
 import org.folio.okapi.util.LogHelper;
 import org.folio.okapi.common.XOkapiHeaders;
@@ -99,7 +98,7 @@ public class MainVerticle extends AbstractVerticle {
     if (loglevel != null) {
       LogHelper.setRootLogLevel(loglevel);
     } else {
-      String lev = getenv("OKAPI_LOGLEVEL");
+      String lev = System.getenv("OKAPI_LOGLEVEL");
       if (lev != null && !lev.isEmpty()) {
         LogHelper.setRootLogLevel(loglevel);
       }
--- a/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
@@ -434,6 +434,7 @@ public class MainVerticle extends AbstractVerticle {
       if (res.succeeded()) {
         logger.info("Deploy completed succesfully");
       } else {
+        // not reporting failure if re-deploy fails
         logger.info("Deploy failed", res.cause());
       }
       if (enableProxy) {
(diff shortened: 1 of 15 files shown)


skoczko:
@adamdickmeiss Can you document changed/additions to the Tenant "init" interface?

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -1108,14 +1092,13 @@ private void authForSystemInterface(ModuleDescriptor authMod, RoutingEntry filt,
...
-        fut.handle(new Failure<>(res.getType(), res.cause()));
+        fut.handle(res);
         return;
       }
       OkapiClient cli = res.result();
       String deftok = cli.getRespHeaders().get(XOkapiHeaders.TOKEN);
-      logger.debug("authForSystemInterface:"
-        + Json.encode(cli.getRespHeaders().entries()));
+      logger.debug("authForSystemInterface: {}",
+        Json.encode(cli.getRespHeaders().entries()));
```
julianladisch:
```suggestion
        () -> Json.encode(cli.getRespHeaders().entries()));
```
Avoid encoding when debug logging is diabled.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -1033,7 +1017,7 @@ private DeploymentDescriptor pickInstance(List<DeploymentDescriptor> instances)
    */
   public void callSystemInterface(Tenant tenant, ModuleInstance inst,
     String request, ProxyContext pc,
-    Handler<ExtendedAsyncResult<OkapiClient>> fut) {
+    Handler<AsyncResult<OkapiClient>> fut) {
```
julianladisch: Good!

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
```diff
@@ -457,7 +434,7 @@ private void checkSuperTenant(String okapiModule, Promise<Void> promise) {
       if (res.succeeded()) {
         logger.info("Deploy completed succesfully");
       } else {
-        logger.info("Deploy failed: {}", res.cause());
+        logger.info("Deploy failed", res.cause());
```
julianladisch: If it is intentional that the failure is only logged and not reported via promise there should be a `//` comment saying so.

okapi-core/src/main/java/org/folio/okapi/managers/PullManager.java
```diff
@@ -52,17 +52,17 @@ private void getRemoteUrl(Iterator<String> it,
     final Buffer body = Buffer.buffer();
     HttpClientRequest req = httpClient.getAbs(url, res1 -> {
       if (res1.failed()) {
-        logger.info("pull for " + baseUrl + " failed with status "
-          + res1.cause().getMessage());
+        logger.warn("pull for {} failed with status {}", baseUrl,
+          res1.cause().getMessage());
```
julianladisch:
```suggestion
        logger.warn("pull for {} failed: {}", baseUrl,
          res1.cause().getMessage(), res1);
```
Add stacktrace.

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
adamdickmeiss: It should probably report failure, but I'd rather not change behavior WRT this, so I'll add a comment.

okapi-core/src/main/java/org/folio/okapi/managers/PullManager.java
adamdickmeiss: There's no stacktrace (not an exception)

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
```diff
@@ -2681,32 +2683,97 @@ private void undeployFirst(Handler<AsyncResult<Void>> fut) {
...
-      conf.remove("mongo_db_init");
-      conf.remove("postgres_db_init");
+      DeploymentOptions opt = new DeploymentOptions().setConfig(conf);
+      vertx.deployVerticle(MainVerticle.class.getName(), opt, res -> {
+        context.assertTrue(res.succeeded());
+        async.complete();
+      });
+    });
+    async.await(1000);
+  }
```
julianladisch: The duplicate code should be extracted into a method:

```
 private void undeployFirstAndDeploy(TestContext context, JsonObject aConf, Handler<AsyncResult<String>> fut) {
   Async myAsync = context.async();
   undeployFirst(context.asyncAssertSuccess(handler -> {
     DeploymentOptions opt = new DeploymentOptions().setConfig(aConf);
     vertx.deployVerticle(MainVerticle.class.getName(), opt, res -> {
       fut.handle(res);
       myAsync.complete();
     });
   }));
   myAsync.await(2000);
 }

 private void undeployFirstAndDeploy(TestContext context, JsonObject aConf) {
   undeployFirstAndDeploy(context, aConf, context.asyncAssertSuccess());
 }
```
Example usage:
```
undeployFirstAndDeploy(context, conf);
```
```
undeployFirstAndDeploy(context, conf, context.asyncAssertFailure(cause -> {
  Assert.assertThat(cause.getMessage(), containsString(
      "password authentication failed for user \"okapi\""));
}));
```

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
julianladisch: Created an issue for this: https://issues.folio.org/browse/OKAPI-816

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
adamdickmeiss: ok. Done

PR #890, opened 2020-03-16
Faster pattern match on routing entries OKAPI-817

adamdickmeiss:


skoczko:
@adamdickmeiss why do we need to keep the regex-based matching? Can't the "fast match" be a replacement?

PR #892, opened 2020-03-19
OkapiToken faster; simpler API OKAPI-818

adamdickmeiss:


skoczko:
Looks good to me.

okapi-common/src/test/java/org/folio/okapi/common/OkapiTokenTest.java
```diff
@@ -1,99 +1,68 @@
...
 
-    ex = false;
-    tok.setToken("");
-    try {
-      String v = tok.getTenant();
-    } catch (IllegalArgumentException e) {
-      ex = true;
-    }
-    assertTrue(ex);
+  private String exceptionMessage(String token) {
```
julianladisch: I've moved the duplicated code into this method and moved each exception test into its own method.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -254,27 +252,24 @@ private String tenantHeader(ProxyContext pc) {
     String tok = ctx.request().getHeader(XOkapiHeaders.TOKEN);
 
     if (auth != null) {
-      Pattern pattern = Pattern.compile("Bearer\\s+(.+)"); // Grab anything after 'Bearer' and whitespace
-      Matcher matcher = pattern.matcher(auth);
-      if (matcher.find() && matcher.groupCount() > 0) {
-        auth = matcher.group(1);
+      if (auth.startsWith("Bearer")) {
```
julianladisch: https://tools.ietf.org/html/rfc6750#section-2.1 requires a single space after Bearer.
Can you check for `startsWith("Bearer ")`?

julianladisch: We don't want to match BearerExtension foo.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -254,27 +252,24 @@ private String tenantHeader(ProxyContext pc) {
...
+        auth = auth.substring(6).trim();
+      }
+      if (tok != null && !auth.equals(tok)) {
+        pc.responseError(400, messages.getMessage("10104"));
+        return null;
+      } else {
```
julianladisch: Can you drop this `else` to reduce indentation?

okapi-common/src/test/java/org/folio/okapi/common/OkapiTokenTest.java
adamdickmeiss: thx

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: sure

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: yep

okapi-common/src/main/java/org/folio/okapi/common/OkapiToken.java
```diff
@@ -15,25 +14,21 @@
...
-  public OkapiToken(RoutingContext ctx) {
-    this.token = ctx.request().getHeader(XOkapiHeaders.TOKEN);
-  }
-
-  public void setToken(String token) {
+  public OkapiToken(String token) {
     this.token = token;
   }
 
   private JsonObject getPayload() {
```
julianladisch: This returns the payload without validating the JWT using the signature. This may be a security issue if used in other contexts.
Can the method been renamed to `getPayloadWithoutValidation` and javadoc added that there is no JWT signature validation?

okapi-common/src/main/java/org/folio/okapi/common/OkapiToken.java
```diff
@@ -15,25 +14,21 @@
...
+      throw new IllegalArgumentException("Missing . separator for token");
+    }
+    idx1++;
+    int idx2 = token.indexOf('.', idx1);
+    if (idx2 == -1) {
+      throw new IllegalArgumentException("Missing . separator for token");
     }
+    String encodedJson = token.substring(idx1, idx2);
     String decodedJson = new String(Base64.getDecoder().decode(encodedJson));
     JsonObject j;
```
julianladisch: You can avoid this variable and save 2 lines by using `return new JsonObject(decodedJson);`

okapi-common/src/main/java/org/folio/okapi/common/OkapiToken.java
adamdickmeiss: ok.. getPayload renamed.. Internal so nothing is affected.

okapi-common/src/main/java/org/folio/okapi/common/OkapiToken.java
adamdickmeiss: yep

PR #894, opened 2020-03-20
Google Code Style plugin and conformance OKAPI-809

adamdickmeiss:
This uses https://maven.apache.org/plugins/maven-checkstyle-plugin/ and https://checkstyle.sourceforge.io/google_style.html to fail the build if there is a code style violation.

Checkstyle's google_style.xml deviates from Google Java Style Guide:
* Checkstyle requires a switch default for enums where Google Java Style Guide doesn't require it: 
https://google.github.io/styleguide/javaguide.html#s4.8.4.3-switch-default
* Indentation of lambdas can be wrong: https://github.com/checkstyle/checkstyle/issues?q=is%3Aopen+label%3Aindentation+lambda
* Checkstyle has many other open issues regarding indentation: https://github.com/checkstyle/checkstyle/labels/indentation 

Suppress false positives using https://checkstyle.sourceforge.io/config_filters.html#SuppressWarningsFilter, for example `@SuppressWarnings("indentation")`.

These classes have been renamed (breaking change!):
* CQLUtil → CqlUtil
* URLDecoder → UrlDecoder


checkstyle-suppressions.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+
+<!DOCTYPE suppressions PUBLIC
+    "-//Puppy Crawl//DTD Suppressions 1.1//EN"
+    "http://checkstyle.sourceforge.net/dtds/suppressions_1_1.dtd">
+
+<suppressions>
+  <suppress checks="AbbreviationAsWordInName"  files="LaunchDescriptor.java"/>
```
julianladisch: Please do not suppress this check for the complete file. Suppress only the three lines where this is needed, for example using `@SuppressWarnings("checkstyle:AbbreviationAsWordInName")` after enabling `<module name="SuppressWarningsHolder" />` and `<module name="SuppressWarningsFilter" />`
as explained on https://checkstyle.sourceforge.io/config_filters.html#SuppressWarningsFilter
The rule shoult catch any names we may add in future.

adamdickmeiss: good idea, I'll try to get that to work

checkstyle.xml
```diff
@@ -0,0 +1,334 @@
...
+
+    <!-- FOLIO: Use Eclipse import order -->
+    <module name="CustomImportOrder">
+      <property name="customImportOrderRules"
+                value="STATIC###STANDARD_JAVA_PACKAGE###SPECIAL_IMPORTS###THIRD_PARTY_PACKAGE"/>
+      <property name="specialImportsRegExp" value="^org\."/>
+      <property name="thirdPartyPackageRegExp" value="^com\."/>
+      <property name="sortImportsInGroupAlphabetically" value="true"/>
+      <property name="separateLineBetweenGroups" value="true"/>
+    </module>
```
julianladisch: Why doesn't this use the suggested default Eclipse configuration from the documentation https://checkstyle.sourceforge.io/config_imports.html#ImportOrder_Examples ?

```
<module name="ImportOrder">
  <property name="groups" value="/^java\./,javax,org,com"/>
  <property name="ordered" value="true"/>
  <property name="separated" value="true"/>
  <property name="option" value="above"/>
  <property name="sortStaticImportsAlphabetically" value="true"/>
</module>
```

checkstyle-suppressions.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0"?>
+
+<!DOCTYPE suppressions PUBLIC
+    "-//Puppy Crawl//DTD Suppressions 1.1//EN"
+    "http://checkstyle.sourceforge.net/dtds/suppressions_1_1.dtd">
+
+<suppressions>
+  <suppress checks="indentation"  files="ProcessModuleHandle.java"/>
```
julianladisch: Please do not suppress the indentation check for the complete file. Suppress only the 3 methods where this is needed, for example using @SuppressWarnings("indentation") after enabling <module name="SuppressWarningsHolder" /> and <module name="SuppressWarningsFilter" />
as explained on https://checkstyle.sourceforge.io/config_filters.html#SuppressWarningsFilter
The rule should still catch any indentation issues we may add in other places of the file in future.

okapi-common/src/main/java/org/folio/okapi/common/CqlUtil.java
```diff
@@ -65,8 +67,8 @@ public static CQLNode reducer(CQLNode vn1, CQLTermNode tn, Comparator<CQLTermNod
       } else {
         CQLSortNode sn = new CQLSortNode(n2);
         List<ModifierSet> mods = n1.getSortIndexes();
-        for (ModifierSet mSet : mods) {
-          sn.addSortIndex(mSet);
+        for (ModifierSet mset : mods) {
+          sn.addSortIndex(mset);
```
julianladisch: This circumvents the intention of the rule to avoid the single letter abbreviation `m` in the first chunk of the two chunk name `mSet` or `mset`. It violates the rule that each word within the variable is camel cased. `mset` is not a word. Either rename to `set` or `modifierSet` or `m` or `s` or use `mods.forEach(sn::addSortIndex)`.

okapi-common/src/main/java/org/folio/okapi/common/CqlUtil.java
```diff
@@ -80,19 +82,22 @@ public static CQLNode reducer(CQLNode vn1, CQLTermNode tn, Comparator<CQLTermNod
...
-  private static CQLNode reduceBoolean(CQLBooleanNode n1, CQLTermNode tn, Comparator<CQLTermNode> cmp) {
+  private static CQLNode reduceBoolean(CQLBooleanNode n1, CQLTermNode tn,
+                                       Comparator<CQLTermNode> cmp) {
+
     CQLNode n2 = null;
     CQLNode left = reducer(n1.getLeftOperand(), tn, cmp);
     CQLNode right = reducer(n1.getRightOperand(), tn, cmp);
 
-    ModifierSet mSet = new ModifierSet(n1.getOperator().toString().toLowerCase());
+    ModifierSet mset = new ModifierSet(n1.getOperator().toString().toLowerCase());
```
julianladisch: Set is a word and must start with a capital letter, see above.

checkstyle.xml
```diff
@@ -0,0 +1,334 @@
...
+                     SR_ASSIGN, STAR, STAR_ASSIGN, LITERAL_ASSERT, TYPE_EXTENSION_AND"/>
+      <message key="ws.notFollowed"
+               value="WhitespaceAround: ''{0}'' is not followed by whitespace. Empty blocks may only be represented as '{}' when not part of a multi-block statement (4.1.3)"/>
+      <message key="ws.notPreceded"
+               value="WhitespaceAround: ''{0}'' is not preceded with whitespace."/>
+    </module>
+    <module name="OneStatementPerLine"/>
+    <module name="MultipleVariableDeclarations"/>
+    <module name="ArrayTypeStyle"/>
+    <module name="MissingSwitchDefault"/>
```
julianladisch: This is more restrictive than Google Code Style, see enum exception: https://google.github.io/styleguide/javaguide.html#s4.8.4.3-switch-default

okapi-common/src/main/java/org/folio/okapi/common/ErrorType.java
```diff
@@ -4,39 +4,36 @@
...
       case FORBIDDEN:
-        code = 403;
-        break;
-      case INTERNAL:
-      case ANY:
-        code = 500;
-        break;
+        return 403;
+      default:
+        return 500;
```
julianladisch: This unnecessary default clause prevents detection of missing enum values if we add more in future.
https://google.github.io/styleguide/javaguide.html#s4.8.4.3-switch-default
Sonar checks for enum completeness: https://sonarcloud.io/organizations/folio-org/rules?open=java%3AS131&rule_key=java%3AS131

okapi-common/src/main/java/org/folio/okapi/common/Messages.java
```diff
@@ -48,12 +49,13 @@
...
     private static final Messages INSTANCE = new Messages();
+
     private SingletonHelper() {
       // prevent instantiation
     }
   }
 
-  /**
+  /** Returns Messages instance (singleton) .
    * @return the singleton instance of Message
```
julianladisch: Remove `@return` line: https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment
Remove space before full stop.

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
```diff
@@ -140,11 +141,11 @@ public String toString() {
   }
 
   /**
-   * static comparsion
+   * Static module comparison.
```
julianladisch: "static" is in the method signature and should not been repeated in the javadoc summary fragment.
It does not compare modules, it compares modul ids.
Can you rename the parameters from i1 to moduleId1 and i2 to moduleId2?
```suggestion
   * Module id comparison.
```

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -580,21 +574,25 @@ private void stopTimer(String tenantId, String moduleId, int seq, Lock lockP) {
...
+  private void handleTimer(String tenantId) {
+    handleTimer(tenantId, null, 0, null);
+  }
+
   private void handleTimer(String tenantId, String moduleId, int seq1, Lock lockP) {
     logger.info("handleTimer tenant {} module {} seq1 {}", tenantId, moduleId, seq1);
-    tenants.get(tenantId, tRes -> {
-      if (tRes.failed()) {
+    tenants.get(tenantId, tres -> {
+      if (tres.failed()) {
```
julianladisch: "tres" has not the correct camel case. You may use res instead.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -580,21 +574,25 @@ private void stopTimer(String tenantId, String moduleId, int seq, Lock lockP) {
...
         // tenant no longer exist
         stopTimer(tenantId, moduleId, seq1, lockP);
         return;
       }
-      Tenant tenant = tRes.result();
-      moduleManager.getEnabledModules(tenant, mRes -> {
-        if (mRes.failed()) {
+      Tenant tenant = tres.result();
+      moduleManager.getEnabledModules(tenant, mres -> {
```
julianladisch: "mres" is not a correct word, wrong camel case. You may use "modRes" instead.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -673,14 +674,14 @@ private void fireTimer(Tenant tenant, ModuleDescriptor md, RoutingEntry re, Stri
     ModuleInstance inst = new ModuleInstance(md, re, path, httpMethod, true);
     MultiMap headers = MultiMap.caseInsensitiveMultiMap();
     logger.info("timer call start module {} for tenant {}", md.getId(), tenantId);
-    proxyService.callSystemInterface("supertenant", headers, tenant, inst, "", cRes -> {
+    proxyService.callSystemInterface("supertenant", headers, tenant, inst, "", cres -> {
```
julianladisch: "cres" is not a correct word, or incorrectly camelCases. You may use "res" instead.

checkstyle.xml
adamdickmeiss: Eclipse might be different from Google Style. The checkstyle.xml here was taken from https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml some time ago.

okapi-common/src/main/java/org/folio/okapi/common/CqlUtil.java
adamdickmeiss: Pedantic. I'm not gonna do it, buf if you make a PR, I'll probably approve

okapi-common/src/main/java/org/folio/okapi/common/CqlUtil.java
adamdickmeiss: Sanme here. Pedantic

checkstyle.xml
adamdickmeiss: OK. I'll guess you'll have to make a PR on the checkstyle.xml that I don't maintain. But I can see that the version I have seems outdated. So I'll try to update to https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml and see what happens

okapi-common/src/main/java/org/folio/okapi/common/ErrorType.java
adamdickmeiss: I got a warning, because no default was added. Pedantic

julianladisch: You may avoid the switch by using an httpCode variable like in the Planet example
https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

checkstyle.xml
julianladisch: Ok, there hasn't been an agreement yet which import order to use: https://issues.folio.org/browse/FOLIO-1667

craigmcnally:
This topic was discussed at the tech leads meeting today - see https://wiki.folio.org/pages/viewpage.action?spaceKey=TLG&title=2020-03-25+Meeting+notes

craigmcnally:
As long as we're consistent I don't really have strong preferences on this sort of thing.  I will say that I like the javadoc comments - do we generate javadocs for OKAPI?

okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
```diff
@@ -177,7 +203,10 @@ public void setMethods(String[] methods) {
     this.methods = methods;
   }
 
-
+  /**
+   * Get path pattern/path - whichever exist.
+   * @return
```
julianladisch: Remove empty `@return`

okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
```diff
@@ -223,6 +252,12 @@ private int skipNamedPattern(String pathPattern, int i, char c) {
     return i;
   }
 
+  /**
+   * set path pattern.
+   * Special constructs like {name} and * are supported.
+   * Throws DecodeException if pattern is invalid.
```
julianladisch:
```suggestion
   * @throws DecodeException if pattern is invalid.
```

okapi-core/src/main/java/org/folio/okapi/managers/DeploymentManager.java
```diff
@@ -62,11 +73,17 @@ public DeploymentManager(Vertx vertx, DiscoveryManager dm, EnvManager em,
...
-    int portEnd = Integer.parseInt(Config.getSysConf("port_end", Integer.toString(portStart + 10), config));
+    int portStart = Integer.parseInt(Config.getSysConf(
+      "port_start", Integer.toString(listenPort + 1), config));
+    int portEnd = Integer.parseInt(Config.getSysConf(
+      "port_end", Integer.toString(portStart + 10), config));
     this.ports = new Ports(portStart, portEnd);
   }
 
+  /**
+   * Unitialize deployment manager.
```
julianladisch: This is contrary to the method name `init` and I don't see it in the code, can you be more verbose?

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
```diff
@@ -77,6 +83,11 @@ public InternalModule(ModuleManager modules, TenantManager tenantManager,
     logger.info("InternalModule starting okapiversion={}", okapiVersion);
   }
 
+  /**
+   * Return module descriptor for okapi itself.
+   * @param okapiVersion Okapi version; null and "0.0.0" will be assumed
```
julianladisch:
```suggestion
   * @param okapiVersion Okapi version; null is converted to "0.0.0"
```

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -160,8 +165,11 @@ public void enableAndDisableCheck(Tenant tenant,
   /**
    * Create a module.
    *
-   * @param md
-   * @param fut
+   * @param md module descriptor
+   * @param check whether to check dependencies
+   * @param preRelease whether to allow pre-release
+   * @param npmSnapshot whehter to allow npm snapshot
```
julianladisch:
```suggestion
   * @param npmSnapshot whether to allow npm snapshot
```

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -90,7 +95,7 @@ private void invalidateCacheEntry(String id) {
   /**
    * Load the modules from the database, if not already loaded.
    *
-   * @param fut
+   * @param fut future
```
julianladisch: This line should be deleted because it does not give any additional information.

checkstyle-suppressions.xml
adamdickmeiss: Done

checkstyle.xml
adamdickmeiss: Using Google style here.

okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
adamdickmeiss: Sure

okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
adamdickmeiss: OK

okapi-core/src/main/java/org/folio/okapi/managers/DeploymentManager.java
adamdickmeiss: A typo. Good catch!

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
adamdickmeiss: null is not converted.

okapi-core/src/main/java/org/folio/okapi/service/impl/ModuleHandleFactory.java
```diff
@@ -12,6 +12,17 @@ private ModuleHandleFactory() {
...
+   * @param vertx Vert.x handle
+   * @param desc launch descriptor
+   * @param id module ID
+   * @param ports ports to be available
+   * @param moduleHost module host override (for Docker)
+   * @param port port to be in use for module
+   * @param config configuration
+   * @return module handle
+   */
   public static ModuleHandle create(Vertx vertx, LaunchDescriptor desc, String id,
```
julianladisch: In addition `id` should be renamed to `tenantId` to also have this information in the method signature. Some IDEs like Eclipse insert the method signature when doing auto-complete.

checkstyle/google_checks.xml
```diff
@@ -0,0 +1,327 @@
+<?xml version="1.0"?>
```
julianladisch: Instead of adding another copy of this file `google_checks.xml` should be moved to `checkstooyle/google_checks.xml`

checkstyle/folio_checks.xml
```diff
@@ -0,0 +1,332 @@
...
+             value="GenericWhitespace ''{0}'' should followed by whitespace."/>
+            <message key="ws.notPreceded"
+             value="GenericWhitespace ''{0}'' is not preceded with whitespace."/>
+        </module>
+        <module name="Indentation">
+            <property name="basicOffset" value="2"/>
+            <property name="braceAdjustment" value="0"/>
+            <property name="caseIndent" value="2"/>
+            <property name="throwsIndent" value="4"/>
+            <property name="lineWrappingIndentation" value="2"/>
```
julianladisch: https://google.github.io/styleguide/javaguide.html#s4.5.2-line-wrapping-indent requires 4.
Continuation lines must have a bigger indentation than the basic offset to make it obvious that they belong to the previous line and are not the first line of the following indented block.

Bad:
```
  private static Boolean checkInterfaceDepAlreadyEnabled(Map<String, ModuleDescriptor> modsEnabled,
    InterfaceDescriptor req) {
    Boolean exist = null;
```

Good:

```
  private static Boolean checkInterfaceDepAlreadyEnabled(Map<String, ModuleDescriptor> modsEnabled,
      InterfaceDescriptor req) {
    Boolean exist = null;
```

adamdickmeiss: I'm okay with that. I was under the impression that Folio was using the +2 and it would be a big effort to change that.

checkstyle/google_checks.xml
adamdickmeiss: I added the "original one" google_checks.xml verbatim as of a couple a days ago to know what the initial version was. Then the folio_checks.xml with the few changes. Yes. google_checks.xml is a copy but from a source that might change. What's the directory "checkstooyle" you are referring to?

checkstyle/folio_checks.xml
adamdickmeiss: OK.. changed to 4.. Don't make me go back! That means the style is now 100% Google except the suppressWarningsFilter / suppressWarningsHolder.

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
julianladisch: This code
```
public static ModuleDescriptor moduleDescriptor(String okapiVersion) {
    String v = okapiVersion;
    if (v == null) {  // happens at compile time,
      v = "0.0.0";   // unit tests can just check for this
```
converts null to "0.0.0".

checkstyle/google_checks.xml
julianladisch: Typo.
Thank you for removing ./google_checks.xml in favour of ./checkstyle/google_checks.xml

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
adamdickmeiss: null is not converted. The version is converted.

okapi-core/src/main/java/org/folio/okapi/service/impl/ModuleHandleFactory.java
adamdickmeiss: Since the id is a module-Id using tenantId here would be very misleading.

pom.xml
```diff
@@ -112,7 +112,39 @@
...
         </configuration>
       </plugin>
-
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>3.1.1</version>
+        <configuration>
+          <configLocation>checkstyle/folio_checks.xml</configLocation>
+          <encoding>UTF-8</encoding>
```
julianladisch: `<linkXRef>false</linkXRef>`
should be inserted to disable this feature that is not properly configured. As suggested by the usage documentation: https://maven.apache.org/plugins/maven-checkstyle-plugin/usage.html#Checking_for_Violations_as_Part_of_the_Build

pom.xml
```diff
@@ -112,7 +112,39 @@
...
+            </goals>
+          </execution>
+          <execution>
+            <!-- report violations -->
+            <id>checkstyle</id>
+            <phase>compile</phase>
+            <goals>
+              <goal>checkstyle</goal>
+            </goals>
+          </execution>
```
julianladisch: This second execution is redundant and should be removed. One can manually run `mvn checkstyle:checkstyle` if needed. The suggested usage does not have a checkstyle goal: https://maven.apache.org/plugins/maven-checkstyle-plugin/usage.html#Checking_for_Violations_as_Part_of_the_Build

pom.xml
adamdickmeiss: thx. will look at that.

pom.xml
adamdickmeiss: ok. I'll see what I can do.

checkstyle.xml
julianladisch: Added deviation from Google style in the issue and pull request description.

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
julianladisch: null is replaced by "0.0.0"?

PR #907, opened 2020-04-06
Adjust test to not fail for insufficient Docker perms OKAPI-823

adamdickmeiss:


okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
```diff
@@ -88,7 +89,6 @@ public void testDockerVersionAtLocal(TestContext context) {
       "mod-users-5.0.0-SNAPSHOT", ports, "localhost", 9232, new JsonObject());
 
     dh.getUrl("/version", res -> {
-      Assume.assumeTrue(res.succeeded());
```
julianladisch, line 91: If you use
* `context.asyncAssertSuccess(res -> {`
* `context.asyncAssertFailure(ex -> {`
* `context.verify(verify -> {`

you can
* avoid manual `async`
* avoid `async.await` 
* avoid `context.assertTrue(res.failed())` and `context.assertTrue(res.succeeded())`
* and use `Assume.assume...` inside an asyncAsserSuccess/asyncAssertFailure/verify block to properly skip a test so that JUnit reports it as skipped.

okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
```diff
@@ -87,28 +80,27 @@ public void testDockerVersionAtLocal(TestContext context) {
...
-              res3.cause().getMessage());
-          }
-          async.complete();
-        });
-      });
+      async.complete();
     });
+    async.await(1000);
+    Assume.assumeTrue(versionRes.containsKey("result"));
+    context.assertTrue(versionRes.getJsonObject("result").containsKey("Platform"));
```
hjiebsco, line 93: Can you remove this line or tweak it? It failed on one of our ec2 box because "Platform" does not exist. The actual result is `{"Version":"17.06.2-ce","ApiVersion":"1.30","MinAPIVersion":"1.12","GitCommit":"402dd4a/17.06.2-ce","GoVersion":"go1.8.3","Os":"linux","Arch":"amd64","KernelVersion":"4.14.94-89.73.amzn2.x86_64","BuildTime":"2017-11-14T22:04:39.163310834+00:00"}`

okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
```diff
@@ -87,28 +80,27 @@ public void testDockerVersionAtLocal(TestContext context) {
...
+    // provoke 404 not found
+    dh.deleteUrl("/version", "msg", context.asyncAssertFailure(cause -> {
+      context.assertTrue(cause.getMessage().startsWith("msg HTTP error 404"),
+        cause.getMessage());
+      // provoke 404 not found
+      dh.postUrlJson("/version", "msg", "{}", context.asyncAssertFailure(cause2 -> {
+        context.assertTrue(cause2.getMessage().startsWith("msg HTTP error 404"),
+          cause2.getMessage());
+      }));
+    }));
```
hjiebsco, line 104: It seems `dh.deleteUrl` and `dh.postUrlJson` can be tested independently. Is there a reason to chain them?

adamdickmeiss: Well. they must be executed only if Docker is available through domain socket. But they could be separate anyway.

adamdickmeiss: It seems to be overkill to move check for docker in Setup.. So I'd rather keep this in one test.

hjiebsco: I did not mean to separate them to different tests. Below is what I was thinking, pure format change to look a bit  nicer.
```
// provoke 404 not found
dh.deleteUrl("/version", "msg", context.asyncAssertFailure(cause -> {
  context.assertTrue(cause.getMessage().startsWith("msg HTTP error 404"), cause.getMessage());
}));
// provoke 404 not found
dh.postUrlJson("/version", "msg", "{}", context.asyncAssertFailure(cause -> {
  context.assertTrue(cause.getMessage().startsWith("msg HTTP error 404"), cause.getMessage());
}));
```

okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
```diff
@@ -87,28 +80,27 @@ public void testDockerVersionAtLocal(TestContext context) {
...
-            context.assertTrue(res3.cause().getMessage().startsWith("msg HTTP error 404"),
-              res3.cause().getMessage());
-          }
-          async.complete();
-        });
-      });
+      async.complete();
     });
+    async.await(1000);
+    Assume.assumeTrue(versionRes.containsKey("result"));
```
julianladisch, line 92: Why don't you use context.verify?

```
    dh.getUrl("/version", res -> context.verify(h -> {
      Assume.assumeTrue(res.succeeded());  // skip test if no access to Okapi
      context.assertTrue(res.result().containsKey("Version"));
      async.complete();
    }));
    async.await(1000);
```

PR #909, opened 2020-04-15
Facility: disable all modules for tenant OKAPI-596

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -269,80 +270,113 @@ public void updateModuleCommit(Tenant t,
...
             return;
           }
-          ead5commit(tenant, mdFrom, mdTo, pc, res1 -> {
-            if (res1.failed()) {
-              fut.handle(new Failure<>(ErrorType.USER, res.cause()));
-              return;
-            }
-            fut.handle(new Success<>(mdTo != null ? mdTo.getId() : ""));
-          });
+          enableAndDisableModule2(tenant, options, mdFrom, mdTo, pc, fut);
```
hjiebsco, line 353: This line can probably be replaced by Line 356 if we remove the else statement.

adamdickmeiss:
> Seems to me that enable and disable modules could be separated to different methods for better readability. :)

Remember that that are three cases. Remove, remove+new =upgrade and, new.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: I don't follow where that is.. The 353 lines .. have else but a lambda.. So I can't fold things together there..

PR #910, opened 2020-04-15
Fix disabling module fails for existing broken deps OKAPI-827

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -138,6 +138,11 @@ void enableAndDisableCheck(Tenant tenant,
       for (ModuleDescriptor md : modlist) {
         mods.put(md.getId(), md);
       }
+      String deps = DepResolution.checkAllDependencies(mods);
+      if (!deps.isEmpty()) {
+        fut.handle(new Success<>()); // failures even before we change enabled modules
```
hjiebsco, line 143: Ignoring pre-existing dependency issue and returning OK are reasonable when deleting modules? Would it be an issue when enabling modules?

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -3031,5 +3031,288 @@ public void testTenantInit(TestContext context) {
 
   }
 
+  @Test
```
hjiebsco, line 3034: Lots of work to set up a test. :)

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
adamdickmeiss: Yes.. How about we just only allow this when removing a module?

PR #913, opened 2020-04-16
Purge and tenantParameters for old-style disable/enable calls OKAPI-831

adamdickmeiss:


okapi-core/src/main/raml/okapi.raml
```diff
@@ -706,6 +715,15 @@ types:
...
+          queryParameters:
+            purge:
+              description: Disabled modules will also be purged.
+              type: boolean
+              required: false
+            tenantParameters:
+              description: Parameters for Tenant init
+              type: string
+              required: false
           body:
```
hjiebsco, line 727: Does this section (mdTo will not be empty, right?) conflict with line 400 `boolean purge = mdTo == null && options.getPurge();`

okapi-core/src/main/raml/okapi.raml
```diff
@@ -666,6 +666,15 @@ types:
...
         description: Disable modules for tenant
+        queryParameters:
+          purge:
+            description: Disabled modules will also be purged.
+            type: boolean
+            required: false
+          tenantParameters:
+            description: Parameters for Tenant init
+            type: string
+            required: false
```
hjiebsco, line 677: Maybe add a purge test for this one too?

okapi-core/src/main/raml/okapi.raml
adamdickmeiss: purge would always be false in this case. mdTo == null ONLY if a module is disabled (no downgrade or upgrade). The purge would be ignored essentially.

okapi-core/src/main/raml/okapi.raml
adamdickmeiss: Could do.. but that would mostly be to check that RAML is ok.. There's 100% coverage.

okapi-core/src/main/raml/okapi.raml
hjiebsco: So this part should be modified because essentially this API does not support "purge" parameter, right?

PR #914, opened 2020-04-17
Okapi 832 invoke flag

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -389,6 +389,10 @@ private void ead1TenantInterface(Tenant tenant, TenantInstallOptions options,
                                    ModuleDescriptor mdFrom, ModuleDescriptor mdTo,
                                    ProxyContext pc, Handler<ExtendedAsyncResult<Void>> fut) {
 
+    if (!options.getInvoke()) {
+      fut.handle(new Success<>());
+      return;
```
hjiebsco, line 394: Make a note that `invoke = false` will skip `purge = true`

adamdickmeiss: Done and I've added a chapter about invoke in the guide

PR #919, opened 2020-04-29
For IntelliJ IDEA use continuation indent of 4

adamdickmeiss:
If IDEA uses editorconfig it uses a contination indent to be same as
indent_size (2).. We want it to be 4. This setting is probably only
recognoized by IDEA.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,3 +16,6 @@ indent_style = tab
 
 [md2toc]
 indent_size = 4
+
+[**.java]
+continuation_indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,6 +6,7 @@ root = true
 charset = utf-8
 end_of_line = lf
 indent_size = 2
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -16,3 +16,6 @@ indent_style = tab
 
 [md2toc]
 indent_size = 4
+
+[**.java]
+continuation_indent_size = 4
```
julianladisch, line 21: If indent_size is defined in [*] then continuation_indent_size should be there as well so that it applies to *.xml, Jenkinsfile, *.sql etc.

julianladisch:
https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties says that `continuation_indent_size` is recognized by Python and JetBrains (IntelliJ IDEA).

.editorconfig
adamdickmeiss: ok. I was trying to be cautious ... Now everywhere

PR #931, opened 2020-05-31
Allow delegate CORS to module

hjiebsco:
See [OKAPI-847](https://issues.folio.org/browse/OKAPI-847): Conditionally defer CORS handling to module when invoked via passthrough API

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
@@ -5,14 +5,11 @@ import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
-import io.vertx.core.http.HttpHeaders;
-import io.vertx.core.http.HttpMethod;
 import io.vertx.core.http.HttpServerOptions;
 import io.vertx.core.json.Json;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.spi.cluster.ClusterManager;
 import io.vertx.ext.web.Router;
-import io.vertx.ext.web.handler.CorsHandler;
 import java.lang.management.ManagementFactory;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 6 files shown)


okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -538,6 +539,18 @@ public void proxy(RoutingContext ctx) {
...
+            ctx.data().put(MainVerticle.DELEGATE_CORS, true);
+            stream.resume();
+            ctx.reroute(ctx.request().path());
+            return;
+          }
+        }
+
         pc.setModList(l);
 
         pc.logRequest(ctx, tenantId);
```
adamdickmeiss, line 556: Please don't import MainVerticle which also includes ProxyService. Can you move all CORS logic to a CorsHandler class? Which is this is then imported/used by ProxyService and MainVerticle.

hjiebsco: Done.

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
```diff
@@ -352,26 +350,9 @@ private void checkSuperTenant(String okapiModule, Promise<Void> promise) {
...
-        .allowedHeader(XOkapiHeaders.REQUEST_ID) //expose response headers
-        .allowedHeader(XOkapiHeaders.MODULE_ID)
-        .exposedHeader(HttpHeaders.LOCATION.toString())
-        .exposedHeader(XOkapiHeaders.TRACE)
-        .exposedHeader(XOkapiHeaders.TOKEN)
-        .exposedHeader(XOkapiHeaders.AUTHORIZATION)
-        .exposedHeader(XOkapiHeaders.REQUEST_ID)
-        .exposedHeader(XOkapiHeaders.MODULE_ID)
-    );
+    CorsHelper.addCorsHanlder(router);
```
craigmcnally, line 355: Typo "Hanlder"

okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
```diff
@@ -0,0 +1,84 @@
...
+
+  private CorsHelper() {
+  }
+
+  /**
+   * Add CORS handler to {@link Router}.
+   *
+   * @param router - {@link Router}
+   */
+  public static void addCorsHanlder(Router router) {
```
craigmcnally, line 32: See above comment about typo

okapi-core/src/main/raml/RoutingEntry.json
```diff
@@ -68,6 +68,10 @@
       "items": {
         "type": "string"
       }
+    },
+    "delegateCORS": {
+      "description": "Okapi handles CORS by default. Set to true to delegate CORS handling to the module",
```
craigmcnally, line 73: consider mentioning that it's only done when using /_/invoke/tenant/<tid>/<path>

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
hjiebsco: Fixed.

okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
hjiebsco: Fixed.

okapi-core/src/main/raml/RoutingEntry.json
hjiebsco: Added.

PR #934, opened 2020-06-05
OKAPI-854: Honor Json-configured postgres_user

julianladisch:


okapi-core/src/main/java/org/folio/okapi/service/impl/PostgresHandle.java
```diff
@@ -57,8 +57,7 @@
         logger.warn("Bad postgres_port value: {}: {}", val, e.getMessage());
       }
     }
-    connectOptions.setUser(Config.getSysConf("postgres_username",
-        Config.getSysConf("postgres_user", "okapi", new JsonObject()), conf));
+    connectOptions.setUser(Config.getSysConf("postgres_username", "postgres_user", "okapi", conf));
```
hjiebsco, line 60: Any reason we need to support both `postgres_username` and `postgres_user`?

okapi-common/src/main/java/org/folio/okapi/common/Config.java
```diff
@@ -26,6 +27,38 @@ public static String getSysConf(String key, String def, JsonObject conf) {
...
+    }
+    v = conf.getString(key1);
+    if (StringUtils.isNotEmpty(v)) {
+      return v;
+    }
+    v = conf.getString(key2);
+    if (StringUtils.isNotEmpty(v)) {
+      return v;
+    }
+    return def;
```
hjiebsco, line 59: Would it be safer to use `StringUtils.isNotBlank()`?

adamdickmeiss: Yes. Let's just honor postgres_username and avoid extra complexity like this. See https://issues.folio.org/browse/OKAPI-798 .. Should never have had support for an alias like that.

okapi-core/src/main/java/org/folio/okapi/service/impl/PostgresHandle.java
adamdickmeiss: No good reason. `postgres_user` should die

julianladisch:
A new pull request for OKAPI-854 will be created that fixes the -D postgres_user usage to -D postgres_username and adds documentation about this.

PR #935, opened 2020-06-09
OKAPI-854: Deprecate postgres_user, switch to postgres_username

julianladisch:
Add deprecation comments to doc/guide.md, PostgresHandle.java and dist/okapi.sh.
Change environment variable postgres_user to postgres_username in dist/okapi.sh
while supporting the deprecated form used by
https://github.com/folio-org/folio-ansible/blob/q4-2019/roles/okapi/templates/okapi.conf.j2
https://github.com/folio-org/okapi-debian/tree/debian-debian/dist


julianladisch:
Yes, but the two subprojects mentioned in the pull request description still use postgres_user. They can only switch to postgres_username when dist/okapi.sh accepts it. This PR changes dist/okapi.sh to accept it.

PR #937, opened 2020-06-10
OKAPI-835 secure APIs is by default

adamdickmeiss:
All end points now have a required permission.


okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
```diff
@@ -216,12 +216,12 @@ public static ModuleDescriptor moduleDescriptor(String okapiVersion) {
         + "   }, {"
         + "    \"methods\" :  [ \"GET\" ],"
         + "    \"pathPattern\" : \"/_/proxy/modules\","
-        + "    \"permissionsRequired\" : [ ], "
+        + "    \"permissionsRequired\" : [ \"okapi.proxy.modules.list\" ], "
```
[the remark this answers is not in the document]
adamdickmeiss: Sure.

adamdickmeiss:
> There are some permission naming inconsistence. For example: `okapi.proxy.tenants.modules.get` vs `okapi.proxy.tenants.interfaces.list`. Since it is a breaking change anyway, maybe it is an opportunity to rename some old permissions to make all follow consistent naming pattern?

I don't think that one is inconsistent.. both have this form when an ID is given for GET:

`okapi.proxy.tenants.modules.get`
`okapi.proxy.tenants.interfaces.get`

and for GET all:

`okapi.proxy.modules.list`
`okapi.proxy.tenants.interfaces.list`

adamdickmeiss:
I am however missing the named permission sets entirely! Will fix that.

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
```diff
@@ -1241,12 +1250,21 @@ public void testSystemInterfaces(TestContext context) {
...
+        .get("/permResult")
+        .then()
+        .statusCode(200)
+        .log().ifValidationFails()
+        .extract().body().asString();
+
+    JsonArray ar = new JsonArray(body);
+    context.assertEquals(2, ar.size());
+    context.assertEquals("okapi-0.0.0", ar.getJsonObject(0).getString("moduleId"));
+    context.assertEquals("header-1", ar.getJsonObject(1).getString("moduleId"));
```
julianladisch, line 1267:
```suggestion
    .body("$", hasSize(2))
    .body("[0].moduleId", is("okapi-0.0.0"))
    .body("[1].moduleId", is("header-1"));
```
RestAssured has JSON path (GPath syntax) support. Simply use a recent RestAssured version, put `RestAssured.defaultParser = Parser.JSON;` into setUp, and import the hamcrest matcher:
```
+import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
+import static org.hamcrest.Matchers.is;
```

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
```diff
@@ -1330,9 +1348,20 @@ public void testSystemInterfaces(TestContext context) {
...
+        .header("X-Okapi-Tenant", okapiTenant)
+        .get("/permResult")
+        .then()
+        .statusCode(200)
+        .log().ifValidationFails()
+        .extract().body().asString();
+
+    ar = new JsonArray(body);
+    context.assertEquals(1, ar.size());
+    context.assertEquals(new JsonObject(expPerms).encode(), ar.getJsonObject(0).encode());
```
julianladisch, line 1363: encode() does not guarantee the order of the keys. You probably want to compare without encode().

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
```diff
@@ -1361,9 +1390,20 @@ public void testSystemInterfaces(TestContext context) {
...
+        .header("X-Okapi-Tenant", okapiTenant)
+        .get("/permResult")
+        .then()
+        .statusCode(200)
+        .log().ifValidationFails()
+        .extract().body().asString();
+
+    ar = new JsonArray(body);
+    context.assertEquals(1, ar.size());
+    context.assertEquals(new JsonObject(expPerms2).encode(), ar.getJsonObject(0).encode());
```
julianladisch, line 1405: encode() does not guarantee the order of the keys. You probably want to compare without encode().

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
```diff
@@ -1420,8 +1460,20 @@ public void testSystemInterfaces(TestContext context) {
...
+        .header("X-Okapi-Tenant", okapiTenant)
+        .get("/permResult")
+        .then()
+        .statusCode(200)
+        .log().ifValidationFails()
+        .extract().body().asString();
+
+    ar = new JsonArray(body);
+    context.assertEquals(2, ar.size());
+    context.assertEquals(new JsonObject(expPerms).encode(), ar.getJsonObject(1).encode());
```
julianladisch, line 1475: encode() does not guarantee the order of the keys. You probably want to compare without encode().

okapi-test-header-module/src/test/java/HeaderModuleTest.java
```diff
@@ -78,11 +81,19 @@ public void test3(TestContext context, Async async) {
...
+    cli.post("/_/tenantPermissions", "{", res1 -> {
+      context.assertTrue(res1.failed());
+      JsonObject perm = new JsonObject("{\"k\": \"v\"}");
+      cli.post("/_/tenantPermissions", perm.encode(), res2 -> {
+        context.assertTrue(res2.succeeded());
+        context.assertEquals("GET test-header-module /_/tenantPermissions 200 -",
+            cli.getRespHeaders().get(XOkapiHeaders.TRACE));
+        cli.get("/permResult", res3 -> {
+          cli.close();
+          context.assertEquals(perm.encode(), new JsonArray(res3.result()).getJsonObject(0).encode());
```
julianladisch, line 93: encode() does not guarantee the order of the keys. You probably want to compare without encode().

adamdickmeiss: yes.

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
adamdickmeiss: yes

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
adamdickmeiss: yes

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
adamdickmeiss: yes

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
adamdickmeiss: yet another way of doing things.. But shorter, so I've done that.

PR #940, opened 2020-06-16
OKAPI-858: okapi.sh should not put credentials into -D command line

julianladisch:
Collect all postgres_* variables, encode them as JSON, write them to
okapi-postgres.conf (at the writeable directory PID_DIR), and pass
the file via the -conf parameter.

Include postgres_server_pem only if defined (even if empty string).

### DIFF
--- a/dist/okapi.sh
+++ b/dist/okapi.sh
@@ -27,6 +27,10 @@ parse_okapi_conf()  {
 
       if [ "$storage" == "postgres" ]; then
          OKAPI_JAVA_OPTS+=" -Dstorage=postgres"
+         OKAPI_OPTIONS+=" -conf ${PID_DIR}/okapi-postgres.conf"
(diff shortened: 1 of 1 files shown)


adamdickmeiss:
> It looks good.. But I can't test it. We need to check that the package that includes utility jq is required in debian/control.

running ansible box reveals this for okapi:
apt-cache show okapi

Depends: openjdk-8-jdk | openjdk-8-jre, adduser, postgresql-client-common

It needs jq in there.

We need access to debian/control. How about just including directory debian as part of okapi. Then the package can be built and tested locally.

julianladisch:
Yes, I support merging https://github.com/folio-org/okapi-debian into https://github.com/folio-org/okapi
jq should be added to debian/control.
folio-ansible already uses it: https://github.com/folio-org/folio-ansible/blob/master/roles/common/tasks/main.yml#L30

PR #944, opened 2020-06-17
Load permissions before tenant init OKAPI-859

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -420,66 +435,82 @@ private void ead1TenantInterface(Tenant tenant, TenantInstallOptions options,
...
+   *
+   * @param tenant tenant
+   * @param options install options
+   * @param mdFrom module from
+   * @param mdTo module to
+   * @param pc proxy context
+   * @param fut response
+   */
+  private void invokePermissions(Tenant tenant, TenantInstallOptions options,
+                                 ModuleDescriptor mdFrom, ModuleDescriptor mdTo,
```
julianladisch, line 456: Can the unused parameter `mdFrom` been removed?

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -375,12 +375,18 @@ private void enableAndDisableModule3(Tenant tenant, TenantInstallOptions options
           fut.handle(new Failure<>(res2.getType(), res2.cause()));
           return;
         }
-        ead5commit(tenant, mdFrom, mdTo, pc, res3 -> {
-          if (res3.failed()) {
-            fut.handle(new Failure<>(ErrorType.USER, res3.cause()));
+        invokePermissionsPermMod(tenant, options, mdFrom, mdTo, pc, res4 -> {
```
julianladisch, line 378: The names of res3 und res4 should be swapped.

PR #948, opened 2020-06-30
OKAPI-866 OOM bulk upload

adamdickmeiss:
This should decrease the memory usage for upload and download for Okapi. The problem is fixed by switching from
https://vertx.io/docs/apidocs/io/vertx/core/http/HttpClientRequest.html#write-io.vertx.core.buffer.Buffer-
to
https://vertx.io/docs/apidocs/io/vertx/core/http/HttpClientRequest.html#write-io.vertx.core.buffer.Buffer-io.vertx.core.Handler-

The problem with the former is that, since it will not block it will pile up the request or response in memory.. Using the latter and pause the source from data until complete minimizes memory.

This PR has a test-case for okapi-test-module.. That is an end-to-end client + server to analyze the problem, which turned out to be useful.

2nd part is "just" the fix for Okapi and a similar test case .

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -744,15 +744,25 @@ public class ProxyService {
   private static void streamHandle(ProxyContext pc, ReadStream<Buffer> readStream,
                                    WriteStream<Buffer> mainWriteStream,
                                    List<HttpClientRequest> logWriteStreams) {
+    List<WriteStream<Buffer>> writeStreams = new LinkedList<>();
+    writeStreams.add(mainWriteStream);
+    for (WriteStream<Buffer> w : logWriteStreams) {
+      writeStreams.add(w);
+    }
+    pumpOneToMany(readStream, writeStreams);
+    readStream.exceptionHandler(e
+        -> pc.warn("streamHandle: content exception ", e));
+    readStream.resume();
+  }
+
+  private static void pumpOneToMany(ReadStream<Buffer> readStream,
+                                    List<WriteStream<Buffer>> writeStreams) {
+
     readStream.handler(data -> {
       AtomicInteger pend = new AtomicInteger();
       // two passes.. to avoid drainHandler being fired off too early.
       // first pass: see if any of writing streams are full?
-      mainWriteStream.write(data);
-      if (mainWriteStream.writeQueueFull()) {
-        pend.incrementAndGet();
-      }
-      for (WriteStream<Buffer> w : logWriteStreams) {
+      for (WriteStream<Buffer> w : writeStreams) {
         w.write(data);
         if (w.writeQueueFull()) {
           pend.incrementAndGet();
@@ -763,14 +773,7 @@ public class ProxyService {
       }
       // second pass: at least one was full. pause and set up drainHandlers for full ones.
       readStream.pause();
-      if (mainWriteStream.writeQueueFull()) {
-        mainWriteStream.drainHandler(x -> {
-          if (pend.decrementAndGet() == 0) {
-            readStream.resume();
-          }
-        });
-      }
-      for (WriteStream<Buffer> w : logWriteStreams) {
+      for (WriteStream<Buffer> w : writeStreams) {
         if (w.writeQueueFull()) {
           w.drainHandler(x -> {
             if (pend.decrementAndGet() == 0) {
@@ -779,17 +782,12 @@ public class ProxyService {
           });
         }
       }
-
     });
     readStream.endHandler(v -> {
-      for (WriteStream<Buffer> w : logWriteStreams) {
+      for (WriteStream<Buffer> w : writeStreams) {
         w.end();
       }
-      mainWriteStream.end();
     });
-    readStream.exceptionHandler(e
-        -> pc.warn("streamHandle: content exception ", e));
-    readStream.resume();
   }
 
   private void proxyRequestResponse(Iterator<ModuleInstance> it,
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -745,19 +745,38 @@ public class ProxyService {
                                    WriteStream<Buffer> mainWriteStream,
                                    List<HttpClientRequest> logWriteStreams) {
     readStream.handler(data -> {
-      readStream.pause();
-      AtomicInteger pend = new AtomicInteger(1 + logWriteStreams.size());
-      mainWriteStream.write(data, x -> {
-        if (pend.decrementAndGet() == 0) {
-          readStream.resume();
-        }
-      });
+      AtomicInteger pend = new AtomicInteger();
+      // two passes.. to avoid drainHandler being fired off too early.
+      // first pass: see if any of writing streams are full?
+      mainWriteStream.write(data);
+      if (mainWriteStream.writeQueueFull()) {
+        pend.incrementAndGet();
+      }
       for (WriteStream<Buffer> w : logWriteStreams) {
-        w.write(data, x -> {
-          if (pend.decrementAndGet() == 0) {
-            readStream.resume();
+        w.write(data);
+        if (w.writeQueueFull()) {
+          pend.incrementAndGet();
+        }
+      }
+      if (pend.get() > 0) {
+        // second pass: at least one was full. pause and set up drainHandlers for full ones.
+        readStream.pause();
+        if (mainWriteStream.writeQueueFull()) {
+          mainWriteStream.drainHandler(x -> {
+            if (pend.decrementAndGet() == 0) {
+              readStream.resume();
+            }
+          });
+        }
+        for (WriteStream<Buffer> w : logWriteStreams) {
+          if (w.writeQueueFull()) {
+            w.drainHandler(x -> {
+              if (pend.decrementAndGet() == 0) {
+                readStream.resume();
+              }
+            });
           }
-        });
+        }
       }
     });
     readStream.endHandler(v -> {
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -592,7 +592,8 @@ public class ProxyService {
         for (HttpClientRequest r : clientRequestList) {
           r.write(data);
         }
-        ctx.response().write(data);
+        res.pause();
+        ctx.response().write(data, end -> res.resume());
       });
       res.endHandler(v -> {
         pc.closeTimer();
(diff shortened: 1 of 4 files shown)


okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -592,7 +592,8 @@ private void proxyResponseImmediate(ProxyContext pc, ReadStream<Buffer> res,
         for (HttpClientRequest r : clientRequestList) {
           r.write(data);
```
julianladisch, line 593: If r can handle less data per second than ctx.response() the data in r's queue will grow until out of memory.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -798,7 +799,8 @@ private void proxyRequestResponse(Iterator<ModuleInstance> it,
       stream.handler(data -> {
         pc.trace("proxyRequestResponse request chunk '"
             + data.toString() + "'");
-        clientRequest.write(data);
+        stream.pause();
+        clientRequest.write(data, comp -> stream.resume());
         for (HttpClientRequest r : clientRequestList) {
           r.write(data);
```
julianladisch, line 805: If r can handle less data per second than clientRequest the data in r's queue will grow until out of memory.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -592,7 +592,8 @@ private void proxyResponseImmediate(ProxyContext pc, ReadStream<Buffer> res,
...
         }
-        ctx.response().write(data);
+        res.pause();
```
julianladisch, line 595: res should be renamed to stream or readStream

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: More theoretical than practical. But I'll change and resume when all writes are done.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: More theoretical than practical. But I'll change and resume when all writes are done.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: Shouldn't be a problem.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -592,7 +592,8 @@ private void proxyResponseImmediate(ProxyContext pc, ReadStream<Buffer> res,
...
+        ctx.response().write(data, end -> res.resume());
```
julianladisch, line 596: write(Buffer, Handler) does NOT wait if the queue is full. It calls the Handler after appending Buffer to the queue regardless of the queue status.
The documenation
* https://vertx.io/docs/apidocs/io/vertx/core/streams/WriteStream.html
* https://vertx.io/docs/vertx-core/java/#streams

explains to call writeQueueFull() to get a signal about a full queue.

You probably need to implement
* https://vertx.io/docs/apidocs/io/vertx/core/streams/Pump.html
* https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/streams/impl/PipeImpl.java

with one ReadStream and multiple WriteStreams.

adamdickmeiss: Yes. That's why it has to pile up in memory if it can't immediately write it.

adamdickmeiss: write(Buffer data, Handler<AsyncResult<Void>> handler)

Same as write(Buffer) but with an handler called when the operation completes

*and* and FWIW, Hongwei sees no memory issues.

adamdickmeiss: I can try to make a solution that does not use write with handler, but with writeQueueFull and drainHandler.

adamdickmeiss: OK. Giving up on that. drainHandler is a NetSocket only thing.

adamdickmeiss: OK.. Now in place.. More complicated than before.. And as far as I can see results are similar, memory wise. Hongwei, you'll have to test this one.

hjiebsco: Tested. Still works.

julianladisch:
```
package tmp;

import org.apache.commons.lang3.StringUtils;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import io.vertx.core.Vertx;
import io.vertx.core.http.HttpClientRequest;
import io.vertx.junit5.VertxExtension;
import io.vertx.junit5.VertxTestContext;

@ExtendWith(VertxExtension.class)
class StreamTest {
  String chunk = StringUtils.repeat("chunk ", 20000);

  void writeChunk(int i, HttpClientRequest request) {
    request.write(chunk, res -> writeChunk(i+1, request));
    System.out.println(i + " " + request.writeQueueFull() + " ");
  }

  @Test()
  void test(Vertx vertx, VertxTestContext vtc) {
    vertx
    .createHttpServer()
    .requestHandler(req -> {
      req.pause();
      // don't fetch the data
    })
    .listen(8888, server -> {
      HttpClientRequest request = vertx
      .createHttpClient()
      .post(8888, "localhost", "/")
      .setChunked(true)
      .handler(handler -> {});
      writeChunk(1, request);
    });
  }
}
package tmp;

import org.apache.commons.lang3.StringUtils;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import io.vertx.core.Vertx;
import io.vertx.core.http.HttpClientRequest;
import io.vertx.junit5.VertxExtension;
import io.vertx.junit5.VertxTestContext;

@ExtendWith(VertxExtension.class)
class StreamTest {
  String chunk = StringUtils.repeat("chunk ", 20000);

  void writeChunk(int i, HttpClientRequest request) {
    request.write(chunk, res -> writeChunk(i+1, request));
    System.out.println(i + " " + request.writeQueueFull() + " ");
  }

  @Test()
  void test(Vertx vertx, VertxTestContext vtc) {
    vertx
    .createHttpServer()
    .requestHandler(req -> {
      req.pause();
      // don't fetch the data
    })
    .listen(8888, server -> {
      HttpClientRequest request = vertx
      .createHttpClient()
      .post(8888, "localhost", "/")
      .setChunked(true)
      .handler(handler -> {});
      writeChunk(1, request);
    });
  }
}
```
Output:
```
1 false 
2 true 
3 true 
4 true 
...
```

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -754,6 +741,57 @@ private void fixupXOkapiToken(ModuleDescriptor md, MultiMap reqHeaders, MultiMap
     }
   }
 
+  private static void streamHandle(ProxyContext pc, ReadStream<Buffer> readStream,
+                                   WriteStream<Buffer> mainWriteStream,
+                                   List<HttpClientRequest> logWriteStreams) {
```
julianladisch, line 746: `List<HttpClientRequest>` can be changed to `List<? extends WriteStream<Buffer>>`

julianladisch: Instead of handling mainWriteStream and logWriteStreams separately they should be combined into one array or list.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -754,6 +741,57 @@ private void fixupXOkapiToken(ModuleDescriptor md, MultiMap reqHeaders, MultiMap
...
+        }
+      }
+      if (pend.get() == 0) {
+        return;
+      }
+      // second pass: at least one was full. pause and set up drainHandlers for full ones.
+      readStream.pause();
+      if (mainWriteStream.writeQueueFull()) {
+        mainWriteStream.drainHandler(x -> {
+          if (pend.decrementAndGet() == 0) {
```
julianladisch, line 768: This won't work in all cases.
A WriteStream will always call the drainHandler, even if ProxyService missed the time period where the WriteStream queue was full.

How about a pauseAndResume method:
```
for each WriteStream
  if not full
    set drainHandler to null
  else
    set drainHandler to pauseAndResume
    pause readStream
    exit this method.
resume readStream
```
Call this method after data has been written to the WriteStreams.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -754,6 +741,55 @@ private void fixupXOkapiToken(ModuleDescriptor md, MultiMap reqHeaders, MultiMap
...
+    readStream.exceptionHandler(e
+        -> pc.warn("streamHandle: content exception ", e));
+    readStream.resume();
+  }
+
+  private static void pumpOneToMany(ReadStream<Buffer> readStream,
+                                    List<WriteStream<Buffer>> writeStreams) {
+
+    readStream.handler(data -> {
+      AtomicInteger pend = new AtomicInteger();
```
julianladisch, line 762: Using an AtomicInteger won't work in all cases:
1. The drainHandler may be called asynchronously at a later time. It may happen that the ProxyService missed the time period where the WriteStream queue was full and only sees the drainHandler call. 
2. A WriteStream implementation may have a high watermark for writeQueueFull() and a low watermark for calling drainHandler. Example: https://github.com/eclipse-vertx/vert.x/blob/3.9.1/src/main/java/io/vertx/core/file/impl/AsyncFileImpl.java#L70-L71 where high watermark is 128 * 1024 and low watermark is half of this = 64 * 1024. 

This algorithm for a pauseAndResume method avoids any counting and only uses the current full status:
```
for each WriteStream
  if not full
    set drainHandler to null
  else
    set drainHandler to pauseAndResume
    pause readStream
    exit this method.
// this point is only reached if all WriteStreams are not full
resume readStream
```
Call this method after data has been written to the WriteStreams.

adamdickmeiss: If you by "exit this method" mean return the for loop entirely.. Then the buffer that is passed in the readStream handler is "lost" for all writers not visited in for loop.

adamdickmeiss:
Thx for looking into this. Didn't realize somebody was working on this branch until I was about to push.. It would be good to know that this was the case so I didn't waste 1-2 hours. Merge?

PR #951, opened 2020-07-08
OKAPI-871 FolioLoggingContext

MikhailFokanov:


julianladisch:
The approach looks good. Sample log entries (they are single lines, I have reformatted them for readability here):
```
{
    "instant": {
        "epochSecond": 1594294449,
        "nanoOfSecond": 725000000
    },
    "thread": "vert.x-eventloop-thread-3",
    "level": "INFO",
    "loggerName": "okapi",
    "message": "090725/test2 RES 404 - okapi No suitable module found for path /test2 for tenant roskilde",
    "endOfBatch": false,
    "loggerFqcn": "org.apache.logging.log4j.spi.AbstractLogger",
    "threadId": 682,
    "threadPriority": 5,
    "userId": "",
    "moduleId": "",
    "requestId": "090725/test2",
    "tenantId": "roskilde"
}
```
```
{
    "instant": {
        "epochSecond": 1594294449,
        "nanoOfSecond": 725000000
    },
    "thread": "vert.x-eventloop-thread-3",
    "level": "ERROR",
    "loggerName": "okapi",
    "message": "HTTP response code=404 msg=No suitable module found for path /test2 for tenant roskilde",
    "endOfBatch": false,
    "loggerFqcn": "org.apache.logging.log4j.spi.AbstractLogger",
    "threadId": 682,
    "threadPriority": 5,
    "userId": "",
    "moduleId": "",
    "requestId": "090725/test2",
    "tenantId": "roskilde"
}
```
```
{
    "instant": {
        "epochSecond": 1594294587,
        "nanoOfSecond": 564000000
    },
    "thread": "vert.x-eventloop-thread-1",
    "level": "WARN",
    "loggerName": "okapi",
    "message": "pull for http://localhost:9233 failed: Connection refused: localhost/127.0.0.1:9233",
    "thrown": {
        "commonElementCount": 0,
        "localizedMessage": "Connection refused: localhost/127.0.0.1:9233",
        "message": "Connection refused: localhost/127.0.0.1:9233",
        "name": "io.netty.channel.AbstractChannel$AnnotatedConnectException",
        "cause": {
            "commonElementCount": 0,
            "localizedMessage": "Connection refused",
            "message": "Connection refused",
            "name": "java.net.ConnectException",
            "extendedStackTrace": "java.net.ConnectException: Connection refused\n\tat sun.nio.ch.SocketChannelImpl.checkConnect(Native Method) ~[?:1.8.0_252]\n\tat sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:714) ~[?:1.8.0_252]\n\tat io.netty.channel.socket.nio.NioSocketChannel.doFinishConnect(NioSocketChannel.java:327) ~[netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.finishConnect(AbstractNioChannel.java:334) [netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:688) [netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:635) [netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:552) [netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:514) [netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.util.concurrent.SingleThreadEventExecutor$6.run(SingleThreadEventExecutor.java:1044) [netty-common-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) [netty-common-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [netty-common-4.1.42.Final.jar:4.1.42.Final]\n\tat java.lang.Thread.run(Thread.java:748) [?:1.8.0_252]\n"
        },
        "extendedStackTrace": "io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:9233\nCaused by: java.net.ConnectException: Connection refused\n\tat sun.nio.ch.SocketChannelImpl.checkConnect(Native Method) ~[?:1.8.0_252]\n\tat sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:714) ~[?:1.8.0_252]\n\tat io.netty.channel.socket.nio.NioSocketChannel.doFinishConnect(NioSocketChannel.java:327) ~[netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.finishConnect(AbstractNioChannel.java:334) [netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:688) [netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:635) [netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:552) [netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:514) [netty-transport-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.util.concurrent.SingleThreadEventExecutor$6.run(SingleThreadEventExecutor.java:1044) [netty-common-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) [netty-common-4.1.42.Final.jar:4.1.42.Final]\n\tat io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [netty-common-4.1.42.Final.jar:4.1.42.Final]\n\tat java.lang.Thread.run(Thread.java:748) [?:1.8.0_252]\n"
    },
    "endOfBatch": false,
    "loggerFqcn": "org.apache.logging.log4j.spi.AbstractLogger",
    "threadId": 1783,
    "threadPriority": 5,
    "userId": "",
    "moduleId": "",
    "requestId": "106202/proxy",
    "tenantId": "supertenant"
}
```

okapi-core/src/main/java/org/folio/okapi/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,61 @@
...
+  */
+  public static void put(String key, String value) {
+    ContextInternal ctx = (ContextInternal) Vertx.currentContext();
+    if (ctx != null) {
+      getContextMap(ctx).put(key, value);
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  private static ConcurrentMap<String, String> getContextMap(ContextInternal ctx) {
```
julianladisch, line 56: I don't think that we need to create yet another context map.
Vert.x already has a map we can easily access with getLocal and putLocal:
https://github.com/eclipse-vertx/vert.x/blob/4.0.0-milestone5/src/main/java/io/vertx/core/impl/ContextImpl.java#L218-L220
https://github.com/eclipse-vertx/vert.x/blob/4.0.0-milestone5/src/main/java/io/vertx/core/impl/AbstractContext.java#L303-L312

okapi-core/src/main/java/org/folio/okapi/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.logging.log4j.core.lookup.StrLookup;
+
+/**
+ * This class should be used for storing context variables
+ * and use them in logging events.
+ */
+@Plugin(name = "FolioLoggingContext", category = StrLookup.CATEGORY)
+public class FolioLoggingContext implements StrLookup {
+
+  public static final String EMPTY_VALUE = "";
```
julianladisch, line 18: If we have traditional (non-JSON) log lines we may need EMPTY_VALUE as placeholder.
For JSON log lines this is not needed. The key should be completely omitted: https://softwareengineering.stackexchange.com/questions/343710/use-empty-string-null-or-remove-empty-property-in-api-request-response

julianladisch: There is no easy way to omit empty or null values.
Returning "" or adding a default value like `property.requestId =` to the log4j2.properties have the same result.

okapi-core/src/main/java/org/folio/okapi/logging/FolioLoggingContext.java
MikhailFokanov: done

okapi-core/src/main/java/org/folio/okapi/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  /**
+  * Lookup value by key. LogEvent isn't used.
+  *
+  * @param key the name of logging variable (e.g. requestId)
+  * @return value for key or *empty string* if there is no such key
+  */
+  public String lookup(LogEvent event, String key) {
+    ContextInternal ctx = (ContextInternal) Vertx.currentContext();
```
julianladisch, line 37: No casting is needed because getLocal is in Context.

okapi-core/src/main/java/org/folio/okapi/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+    return EMPTY_VALUE;
+  }
+
+  /**
+  * Put value by key to the logging context.   *
+  * @param key the name of logging variable (e.g. requestId)
+  */
+  public static void put(String key, String value) {
+    ContextInternal ctx = (ContextInternal) Vertx.currentContext();
```
julianladisch, line 52: No casting is needed because putLocal is in Context.

okapi-core/src/main/resources/log4j2.properties
```diff
@@ -10,8 +10,27 @@ appenders = console
...
+appender.console.layout.tenantId.key = tenantId
+appender.console.layout.tenantId.value = $${FolioLoggingContext:tenantId}
+
+appender.console.layout.userId.type = KeyValuePair
+appender.console.layout.userId.key = userId
+appender.console.layout.userId.value = $${FolioLoggingContext:userId}
+
+appender.console.layout.moduleId.type = KeyValuePair
+appender.console.layout.moduleId.key = moduleId
+appender.console.layout.moduleId.value = $${FolioLoggingContext:moduleId}
```
julianladisch, line 33: In https://stackoverflow.com/questions/31342950 they advise not to use camel case for plugin names.

okapi-core/src/main/java/org/folio/okapi/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,58 @@
...
+    if (ctx != null) {
+      ctx.putLocal(key, value);
```
julianladisch, line 54: Do we need to add LOGGING_VAR_PREFIX here or should ProxyService prepend it?

MikhailFokanov: I want to make the usage of the approach as simple, as it can be. If for putting and getting every value there was a need to add a prefix, someone would forget to put it on get or put. So I'd like to stick with the approach of putting it internally.

okapi-core/src/main/resources/log4j2.properties
MikhailFokanov: I haven't faced such issues. Maybe they are for old version of log4j2. Are you proposing to lowercase the variables?

julianladisch: It is easier to simply lowercase requestId, tenantId, userId, moduleId and FolioLoggingContext than investigate whether that bug may affect us.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -514,6 +515,18 @@ public void proxy(RoutingContext ctx) {
...
+
+      FolioLoggingContext.put("moduleId", ctx.request().headers().get(XOkapiHeaders.MODULE_ID));
+
+      FolioLoggingContext.put("userId", ctx.request().headers().get(XOkapiHeaders.USER_ID));
+
+    }
+
     final MultiMap headers = ctx.request().headers();
     sanitizeAuthHeaders(headers);
     tenantManager.get(tenantId, gres -> {
```
adamdickmeiss, line 532: Redundant checks here.. headers are always present.  Move your stuff below the headers local and have it more readable, please.

okapi-core/src/main/java/org/folio/okapi/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,62 @@
+package org.folio.okapi.logging;
```
julianladisch, line 1: The class should be moved from okapi-core to okapi-common that already is a dependency of RMB: https://github.com/folio-org/raml-module-builder/blob/v30.2.4/pom.xml#L162

okapi-core/src/main/java/org/folio/okapi/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.config.plugins.Plugin;
+import org.apache.logging.log4j.core.lookup.StrLookup;
+
+/**
+ * This class should be used for storing context variables
+ * and use them in logging events.
+ */
+@Plugin(name = "FolioLoggingContext", category = StrLookup.CATEGORY)
+public class FolioLoggingContext implements StrLookup {
```
julianladisch, line 14: "Folio" can be removed from the class name because "folio" is already in the package name: "org.folio.okapi.logging" or "org.folio.okapi.common"

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
MikhailFokanov: done

okapi-core/src/main/java/org/folio/okapi/logging/FolioLoggingContext.java
MikhailFokanov: done

okapi-core/src/main/java/org/folio/okapi/logging/FolioLoggingContext.java
MikhailFokanov: In log4j2.properties the name "FolioLoggingContext" used as: $${FolioLoggingContext:userid}. If it is just $${LoggingContext:userid}, the name is not self-explaining and there can be some collisions with some other _LoggingContext_

MikhailFokanov: Having 
`
@Plugin(name = "FolioLoggingContext", category = StrLookup.CATEGORY)
public class LoggingContext implements StrLookup {
`
for me also looks not so good.

okapi-core/src/main/resources/log4j2.properties
MikhailFokanov: done

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,70 @@
...
+
+import io.vertx.core.Context;
+import io.vertx.core.Vertx;
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.config.plugins.Plugin;
+import org.apache.logging.log4j.core.lookup.StrLookup;
+
+/**
+ * This class should be used for storing context variables
+ * and use them in logging events.
```
julianladisch, line 11: Can you be more specific about the usage of this class? It should mention log4j, Vertx context, Vertx vertx local data.

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,70 @@
...
+        return val;
+      }
+    }
+    return EMPTY_VALUE;
+  }
+
+  /**
+  * Put value by key to the logging context.   *
+  * @param key the name of logging variable (e.g. requestId)
+  */
```
julianladisch, line 58: Can you add javadoc for value and explain that null is accepted and will remove the entry?

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.config.plugins.Plugin;
+import org.apache.logging.log4j.core.lookup.StrLookup;
+
+/**
+ * This class should be used for storing context variables
+ * and use them in logging events.
+ */
+@Plugin(name = "FolioLoggingContext", category = StrLookup.CATEGORY)
+public class FolioLoggingContext implements StrLookup {
```
julianladisch, line 14: VertxLoggingContext might be a better name.

okapi-common/pom.xml
```diff
@@ -15,7 +15,6 @@
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
       <artifactId>log4j-core</artifactId>
-      <scope>test</scope>
```
julianladisch, line 18: Why isn't this `<optional>true</optional>` to allow users of okapi-common to use some other logging library if they want?

okapi-common/src/test/java/org/folio/okapi/common/FolioLoggingContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @After
+  public void tearDown() {
+    vertx.close();
+  }
+
+  @Test
+  public void lookupWithoutContextTest() {
+    FolioLoggingContext loggingContext = new FolioLoggingContext();
+    loggingContext.lookup(null);
```
julianladisch, line 36: The return value should be asserted.

okapi-common/src/test/java/org/folio/okapi/common/FolioLoggingContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+
+  @Test
+  public void lookupPutTest(TestContext context) {
+    vertx.runOnContext(e -> {
+      FolioLoggingContext loggingContext = new FolioLoggingContext();
+      FolioLoggingContext.put(KEY, VALUE);
+      context.assertEquals(loggingContext.lookup(KEY), VALUE);
```
julianladisch, line 45: This line should be wrapped in yet another vertx.runOnContext to make sure that the entry really is passed on.

okapi-common/src/test/java/org/folio/okapi/common/FolioLoggingContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+  @Test
+  public void lookupWithoutContextTest() {
+    FolioLoggingContext loggingContext = new FolioLoggingContext();
+    loggingContext.lookup(null);
+  }
+
+
+  @Test
+  public void lookupPutTest(TestContext context) {
+    vertx.runOnContext(e -> {
```
julianladisch, line 42: Async is required, otherwise the code inside the runOnContext is executed after test completion and therefore cannot report any failures.

okapi-common/src/test/java/org/folio/okapi/common/FolioLoggingContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+      FolioLoggingContext loggingContext = new FolioLoggingContext();
+      FolioLoggingContext.put(KEY, VALUE);
+      context.assertEquals(loggingContext.lookup(KEY), VALUE);
+    });
+  }
+
+  @Test
+  public void lookupNullTest(TestContext context) {
+    FolioLoggingContext loggingContext = new FolioLoggingContext();
```
julianladisch, line 51: Async is missing

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private static final String LOGGING_VAR_PREFIX = "folio_";
+
+  public static final String TENANT_ID_LOGGING_VAR_NAME = "tenantid";
+
+  public static final String REQUEST_ID_LOGGING_VAR_NAME = "requestid";
+
+  public static final String MODULE_ID_LOGGING_VAR_NAME = "moduleid";
+
+  public static final String USER_ID_LOGGING_VAR_NAME = "userid";
```
julianladisch, line 26: These should be moved to okapi-core ProxyService, because this logging context is a general purpose library class.

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Lookup value by key.
+   *
+   * @param key the name of logging variable (e.g. requestId)
+   * @return value for key or *empty string* if there is no such key
+   */
+  public String lookup(String key) {
```
julianladisch, line 34: Can you add `@Override`?

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,70 @@
...
+    return lookup(null, key);
+  }
+
+  /**
+  * Lookup value by key. LogEvent isn't used.
+  *
+  * @param key the name of logging variable (e.g. requestId)
+  * @return value for key or *empty string* if there is no such key
+  */
+  public String lookup(LogEvent event, String key) {
```
julianladisch, line 44: Can you add `@Override`?

okapi-common/src/test/java/org/folio/okapi/common/FolioLoggingContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+    vertx.runOnContext(e -> {
+      loggingContext.lookup(null);
```
julianladisch, line 53: The return value should be asserted.

okapi-common/src/test/java/org/folio/okapi/common/FolioLoggingContextTest.java
julianladisch, line 36: null as a key may cause an exception, can you use a proper key?

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,70 @@
...
+  public static final String REQUEST_ID_LOGGING_VAR_NAME = "requestid";
+
+  public static final String MODULE_ID_LOGGING_VAR_NAME = "moduleid";
+
+  public static final String USER_ID_LOGGING_VAR_NAME = "userid";
+
+  /**
+   * Lookup value by key.
+   *
+   * @param key the name of logging variable (e.g. requestId)
```
julianladisch, line 31: Can you add whether null key is allowed?
The same for lookup(LogEvent, String)?

okapi-common/src/test/java/org/folio/okapi/common/FolioLoggingContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+    });
+  }
+
+  @Test
+  public void putNullTest(TestContext context) {
+    Async async = context.async();
+    FolioLoggingContext loggingContext = new FolioLoggingContext();
+    vertx.runOnContext(e -> {
+          FolioLoggingContext.put(KEY, null);
+          context.assertEquals(loggingContext.lookup(KEY), "");
```
julianladisch, line 63: You need to swap the order for `context.assertEquals(expected, actual)`.

okapi-common/src/test/java/org/folio/okapi/common/FolioLoggingContextTest.java
julianladisch, line 45: You need to swap the order for `context.assertEquals(expected, actual)`.

okapi-common/src/test/java/org/folio/okapi/common/FolioLoggingContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @Before
+  public void setup() {
+    vertx = Vertx.vertx();
+  }
+
+  @After
+  public void tearDown() {
+    vertx.close();
+  }
```
julianladisch, line 31: Vertx vertx, setup() and tearDown() are not needed and should be removed.

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
MikhailFokanov: These are general platform wide constants, which make sense for every module.

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
MikhailFokanov: I think, that there is a lot of stuff related to Vertx and VertxLoggingContext can be jumble with other Vertx classes. Does that make sense?

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * This class should be used for storing context variables
+ * and use them in logging events.
+ * The variables are stored in the local vert.x
+ * {@link io.vertx.core.Context#putLocal Context}
+ * and can be used in log4j log lines by using
+ * {@link org.apache.logging.log4j.core.lookup.StrLookup StrLookup}
+ *
+ * , Vertx vertx local data
+ */
```
julianladisch, line 18: Can you fix the javadoc? The build fails with a Checkstyle violation.

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,83 @@
...
+        return val;
+      }
+    }
+    return EMPTY_VALUE;
+  }
+
+  /**
+  * Put value by key to the logging context.
+  * @param key the name of logging variable, {@code null} key isn't allowed.
+  * @param value the value of  logging variable.
```
julianladisch, line 69:
```suggestion
  * @param value the value of logging variable.
```

okapi-core/src/main/resources/log4j2-dev.properties
```diff
@@ -0,0 +1,18 @@
+status = error
```
julianladisch, line 1: Adam said that production default should be JSONLayout and development default should be PatternLayout for readability.
Can you copy okapi-core/src/main/resources/log4j2-dev.properties to  okapi-core/src/test/resources/log4j2-test.properties?

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -515,6 +516,16 @@ public void proxy(RoutingContext ctx) {
...
     final MultiMap headers = ctx.request().headers();
+
+    FolioLoggingContext.put(FolioLoggingContext.TENANT_ID_LOGGING_VAR_NAME,
+        tenantId);
+    FolioLoggingContext.put(FolioLoggingContext.REQUEST_ID_LOGGING_VAR_NAME,
+        headers.get(XOkapiHeaders.REQUEST_ID));
+    FolioLoggingContext.put(FolioLoggingContext.MODULE_ID_LOGGING_VAR_NAME,
+        headers.get(XOkapiHeaders.MODULE_ID));
+    FolioLoggingContext.put(FolioLoggingContext.USER_ID_LOGGING_VAR_NAME,
+        headers.get(XOkapiHeaders.USER_ID));
```
julianladisch, line 527: When running `mvn clean install` there isn't a single unit test that logs a user id. Can you extend an existing test by adding XOkapiHeaders.USER_ID? You don't need to assert that it is logged, I can look at the output.

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.logging.log4j.core.config.plugins.Plugin;
+import org.apache.logging.log4j.core.lookup.StrLookup;
+
+/**
+ * This class should be used for storing context variables
+ * and use them in logging events.
+ * The variables are stored in the local vert.x
+ * {@link io.vertx.core.Context#putLocal Context}
+ * and can be used in log4j log lines by using
+ * {@link org.apache.logging.log4j.core.lookup.StrLookup StrLookup}
```
julianladisch, line 15: I don't see StrLookup in log4j2.properties. I only see references like `$${FolioLoggingContext:requestid}`. Can you make the usage explanation more instructive?

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
MikhailFokanov: done

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
MikhailFokanov: done

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
MikhailFokanov: done

okapi-common/src/main/java/org/folio/okapi/common/logging/FolioLoggingContext.java
MikhailFokanov: done

julianladisch, line 15:
```suggestion
 * <code>$${FolioLoggingContext:requestid}</code>
```

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
julianladisch: This works if the user id is passed in as X-Okapi-User-Id header.
However, the front-end passes in the user id in the token. Okapi sends the token to mod-authtoken for validation. When Okapi gets the reply from mod-authtoken that the validation was successful we know the user id and can set the user id in the FolioLoggingContext. Details: https://github.com/folio-org/okapi/blob/master/doc/security.md
Can you find the code where Okapi gets the validation result and add code for setting the user id?

hjiebsco: Are we interested in userId, username (login) or both? I searched Okapi code and looks like currently it does not have a central place to extract those info from the token. I am thinking that we can add those methods to https://github.com/folio-org/okapi/blob/master/okapi-common/src/main/java/org/folio/okapi/common/OkapiToken.java. For example like this https://github.com/folio-org/okapi/compare/username-userid?expand=1

okapi-common/src/test/java/org/folio/okapi/common/FolioLoggingContextTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+  @Test
+  public void lookupWithoutContextTest(TestContext context) {
+    FolioLoggingContext loggingContext = new FolioLoggingContext();
+    context.assertEquals(EMPTY_STRING, loggingContext.lookup(KEY));
+  }
```
hjiebsco, line 39: Maybe also add following lines?
```
    FolioLoggingContext.put(KEY, VALUE);
    context.assertEquals(EMPTY_STRING, loggingContext.lookup(KEY));
```

MikhailFokanov: done

okapi-test-auth-module/src/main/java/org/folio/okapi/auth/Auth.java
```diff
@@ -222,6 +223,8 @@ public void check(RoutingContext ctx) {
     }
     // Fake some module tokens
     String modTok = moduleTokens(ctx);
+    FolioLoggingContext.put(FolioLoggingContext.USER_ID_LOGGING_VAR_NAME,
+        userId);
```
hjiebsco, line 227: As chatted, we do not need to change okapi-test-auth-module/src/main/java/org/folio/okapi/auth/Auth.java

okapi-common/src/main/java/org/folio/okapi/common/OkapiToken.java
```diff
@@ -41,16 +46,37 @@ private JsonObject getPayloadWithoutValidation() {
...
+    }
+    return payloadWithoutValidation.getString(field);
+  }
+
   /**
    * Get the tenant out from the token.
    * Note there is no JWT validation taking place.
    * @return null if no token, or no tenant there
    */
   public String getTenant() {
```
julianladisch, line 61: Taking a value out of a JWT token without validating the token can cause a security issue because anyone can spoof it.
Such a value must be handled with caution and care.
I suggest to rename the methods to draw developer attention to the missing validation:
* rename `getTenant()` to `getTenantWithoutValidation()`
* rename `getUsername()` to `getUsernameWithoutValidation()`
* rename `getUserId()` to `getUserIdWithoutValidation()`
* rename `getFieldFromToken(String)` to `getFieldFromTokenWithoutValidation(String)`

MikhailFokanov: fixed

julianladisch (approved):
Looks good, thanks!

adamdickmeiss (approved):
I would have voted against suffix withoutValidation..for OkapiToken.. Also I don't like that Okapi should know about the userid.. But I guess that's what Log people want.. Okapi until know only knew about tenant part of token .. rest was opaque .. so we have duplicated and somewhat duplicated auth logic with this in okapi. But I understand it's for the sake of logging.

skoczko:
@MikhailFokanov I am a little concerned that the JSON layout is now the default configuration. Can we have a switch for that and use (at least for now) the pattern layout one?

PR #957, opened 2020-08-05
Okapi 865 upgrade to 4 Beta 1

adamdickmeiss:


craigmcnally:
As a convenience to other reviewers:  [https://vertx.io/blog/eclipse-vert-x-4-beta-1-released/]

okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
```diff
@@ -195,11 +194,6 @@ public String getPhaseLevel() {
    * @param methods HTTP method name or "*" for all
    */
   public void setMethods(String[] methods) {
-    for (String s : methods) {
```
craigmcnally, line 198: what is the reason for stripping this out?

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -609,53 +608,55 @@ private void proxyRequestHttpClient(
     RoutingContext ctx = pc.getCtx();
     String url = makeUrl(mi, ctx);
     HttpMethod meth = ctx.request().method();
-    HttpClientRequest clientRequest = httpClient.requestAbs(meth, url, res1 -> {
-      if (proxyHttpFail(pc, mi, res1)) {
```
craigmcnally, line 613: this refactoring is annoying, but at least this is cleaner...

okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
```diff
@@ -99,48 +100,62 @@ static String setupDockerAddress(StringBuilder socketAddress, String u) {
...
                          Handler<AsyncResult<Void>> future) {
+    if (res.failed()) {
+      future.handle(res.mapEmpty());
+      return;
+    }
     Buffer body = Buffer.buffer();
-    res.handler(body::appendBuffer);
-    res.endHandler(d -> {
-      if (res.statusCode() == 204) {
+    res.result().handler(body::appendBuffer);
```
craigmcnally, line 110: why not just call res.result() once?

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -65,10 +69,19 @@
   private int timerDelaySum = 0;
   private int timerTenantInitStatus = 200;
   private int timerTenantPermissionsStatus = 200;
+  private HttpServer listenTimer;
   private JsonObject timerPermissions = new JsonObject();
   private JsonArray edgePermissionsAtInit = null;
   private JsonObject timerTenantData;
 
+  @Rule
+  public TestWatcher watchman = new TestWatcher() {
```
craigmcnally, line 78: neat trick

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -314,14 +342,15 @@ private void td(TestContext context, Async async) {
...
-    HttpClientLegacy.delete(httpClient, port, "localhost", "/_/discovery/modules", response -> {
-      context.assertTrue(response.statusCode() == 404 || response.statusCode() == 204);
-      response.endHandler(x -> {
-        httpClient.close();
-        td(context, async);
-      });
-    }).end();
+    httpClient.delete(port, "localhost", "/_/discovery/modules",
+        MultiMap.caseInsensitiveMultiMap(), res1 -> {
+          context.assertTrue(res1.succeeded());
```
craigmcnally, line 347: not sure it's a big deal, but is this really an equivalent test to 404 or 204? <thinking>

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -2177,20 +2214,20 @@ public void testRequestOnly(TestContext context) {
...
-      + "    \"exec\" : "
-      + "\"java -Dport=%p -jar ../okapi-test-module/target/okapi-test-module-fat.jar\"" + LS
-      + "  }" + LS
-      + "}";
+        + "  \"id\" : \"request-only-1.0.0\"," + LS
+        + "  \"name\" : \"request-only\"," + LS
+        + "  \"filters\" : [ {" + LS
+        + "    \"methods\" : [ \"GET\", \"POST\" ]," + LS
+        + "    \"path\" : \"/testb\"," + LS
+        + "    \"level\" : \"30\"," + LS
```
craigmcnally, line 2222: Just curious why?

okapi-core/src/test/java/org/folio/okapi/bean/BeanTest.java
```diff
@@ -233,7 +233,7 @@ public void testModuleDescriptorBadMethod() {
       + "    \"id\" : \"sample\"," + LS
       + "    \"version\" : \"1.0\"," + LS
       + "    \"handlers\" : [ {" + LS
-      + "      \"methods\" : [ \"GET\", \"NOST\" ]," + LS
+      + "      \"methods1\" : [ \"GET\", \"POST\" ]," + LS
```
craigmcnally, line 236: is having an invalid property value checked elsewhere?

okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
adamdickmeiss: Because it never fires.. In fact HttpMethod.valueOf("GYF)".name() == "GYF".

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: Yes.. It is a better API..

okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
adamdickmeiss: Sure. Will do

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
adamdickmeiss: Adding a test for 204 there.

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
adamdickmeiss: It is meant to be executed after auth (level=10) and and before pre(level=40).

okapi-core/src/test/java/org/folio/okapi/bean/BeanTest.java
adamdickmeiss: Adding one for RoutingEntryTest .. However HttpMethod accepts any value, so Okapi doesn not reject, say, method=GYF

adamdickmeiss: I reverted this and fixed setMethod to throw exception for bad method.

okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
adamdickmeiss: I've fixed this so that exception is thrown for invalid method.

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
craigmcnally: I get that, but by change from 33 to 30?

PR #958, opened 2020-08-11
Okapi 860 metrics

hjiebsco:
Please see [OKAPI-860](https://issues.folio.org/browse/OKAPI-860). Following meters (and tags) are defined and implemented for Okapi HTTP proxy calls. There is no error meter defined for server side because that is covered by server processingTime meter. The naming convention follows [Vert.x core tools metrics](https://vertx.io/docs/vertx-micrometer-metrics/java/#_vert_x_core_tools_metrics). Note, the dot in the name is converted to _ in InfluxDB automatically.

- org.folio.okapi.http.server.processingTime - metrics type is Micrometer Timer. Tags are below
  - host - Okapi instance. An example value: ip-172-31-19-200.ec2.internal/172.31.19.200
  - tenant - FOLIO tenant id. An example value: diku
  - code - HTTP response code. An example value: 200
  - method - HTTP request method. An example value: GET
  - module - FOLIO module id. An example value: mod-authtoken-2.6.0-SNAPSHOT.73
  - url - HTTP request url as defined in module descriptor. An example value: /users/ {id}
- org.folio.okapi.http.client.responseTime - metrics type is Micrometer Timer. Tags are below
  - all tags defined above plus
  - phase - FOLIO proxy phase. An example value: auth. The default value is handler
- org.folio.okapi.http.client.errors - metrics type is Micrometer Counter. Tags are below
  - tags: host, tenant, method, and url

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/util/MetricsHelper.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/MetricsHelper.java
@@ -73,7 +73,7 @@ public class MetricsHelper {
     if (influxUserName != null) {
       influxDbOptions.setUserName(influxUserName);
     }
-    logger.log(Level.INFO, influxDbOptions.toJson().encodePrettily());
+    logger.info("Influx config: {}", influxDbOptions.toJson().encodePrettily());
     if (influxPassword != null) {
       influxDbOptions.setPassword(influxPassword);
     }
--- a/okapi-core/src/main/java/org/folio/okapi/util/MetricsHelper.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/MetricsHelper.java
@@ -15,6 +15,7 @@ import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.Logger;
 import org.folio.okapi.bean.ModuleInstance;
 import org.folio.okapi.common.OkapiLogger;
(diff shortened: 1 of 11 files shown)


okapi-core/src/main/java/org/folio/okapi/util/MetricsHelper.java
```diff
@@ -0,0 +1,216 @@
...
+   * @param influxUrl      - default to http://localhost:8086
+   * @param influxDbName   - default to okapi
+   * @param influxUserName - default to null
+   * @param influxPassword - default to null
+   */
+  public static void config(VertxOptions vertxOptions, String influxUrl,
+      String influxDbName, String influxUserName, String influxPassword) {
+    VertxInfluxDbOptions influxDbOptions = new VertxInfluxDbOptions()
+        .setEnabled(true)
+        .setUri(Optional.ofNullable(influxUrl).orElse("http://localhost:8086"))
```
julianladisch, line 70: There is no need to create an Optional for the sole purpose of providing a default value for null.
There exist several idioms for this:
* String.toString(influxUrl, "http://localhost:8086")
* Object.toString(influxUrl, "http://localhost:8086")
* StringUtils.defaultString(influxUrl, "http://localhost:8086")
* ObjectUtils.firstNonNull(influxUrl, "http://localhost:8086")

(There are 5 such Optional instances in this file.)

hjiebsco: Thanks for the suggestions. For the four simple string assignments, I switched to use ternary operator to check null. That seems more intuitive. I kept Optional for the registry assignment so we do not have to depend on 3rd party libraries. I am not familiar with the first item on your list `String.toString(a, b)`, which utility library provides that?

PR #970, opened 2020-09-11
Okapi 887 unit tests sporadically fails

adamdickmeiss:


craigmcnally:
This helps - the build/tests complete now.  However, I still see a dumpfile being generated - not sure how urgent it is to fix this - I believe it has to do with the version of the surefire plugin we're using.  I've attached the latest dumpfile to the JIRA.

craigmcnally (approved):

PR #971, opened 2020-09-12
Update to Vert.x 4.0.0 Beta 2 OKAPI-889

adamdickmeiss:
Also netty to 4.1.52.Final. There are code changes due to the
methods that have been removed from HttpClient.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/DiscoveryManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/DiscoveryManager.java
@@ -566,14 +566,16 @@ public class DiscoveryManager implements NodeListener {
     futures.all(all, fut);
   }
 
-  void fail(Throwable cause, HealthDescriptor hd, Handler<ExtendedAsyncResult<HealthDescriptor>> fut) {
+  void fail(Throwable cause, HealthDescriptor hd,
+            Handler<ExtendedAsyncResult<HealthDescriptor>> fut) {
+
     hd.setHealthMessage("Fail: " + cause.getMessage());
     hd.setHealthStatus(false);
     fut.handle(new Success<>(hd));
   }
 
   void health(DeploymentDescriptor md,
-                      Handler<ExtendedAsyncResult<HealthDescriptor>> fut) {
+              Handler<ExtendedAsyncResult<HealthDescriptor>> fut) {
 
     HealthDescriptor hd = new HealthDescriptor();
     String url = md.getUrl();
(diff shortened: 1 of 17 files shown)


okapi-core/src/main/java/org/folio/okapi/managers/DiscoveryManager.java
```diff
@@ -578,23 +587,28 @@ void health(DeploymentDescriptor md,
...
+        HttpClientResponse response = res.result();
+        response.endHandler(x -> {
+          hd.setHealthMessage("OK");
+          hd.setHealthStatus(true);
+          fut.handle(new Success<>(hd));
+        });
+        response.exceptionHandler(x -> {
+          hd.setHealthMessage("Fail: " + x.getMessage());
+          hd.setHealthStatus(false);
+          fut.handle(new Success<>(hd));
```
julianladisch, line 610: x should be renamed to e and the fail method been used.

okapi-core/src/test/java/org/folio/okapi/MultiTenantTest.java
```diff
@@ -142,15 +143,19 @@ private void td(TestContext context, Async async) {
...
+    httpClient.request(HttpMethod.DELETE, port,
+        "localhost", "/_/discovery/modules", context.asyncAssertSuccess(request -> {
+          request.end();
+          request.onComplete(context.asyncAssertSuccess(response -> {
+            context.assertEquals(204, response.statusCode());
+            response.endHandler(x -> {
+              httpClient.close();
+              vertx.close(y -> {
+                async.complete();
+              });
```
julianladisch, line 155: This is a more natural way because it also checks that closing vertx succeeds:

```
httpClient.close();
vertx.close(context.asyncAssertSuccess());
async.complete();
```

okapi-core/src/test/java/org/folio/okapi/OkapiPerformance.java
```diff
@@ -55,53 +56,11 @@ public void setUp(TestContext context) {
...
-        });
-      }));
-      return;
-    }
-    vertx.close(x -> {
-      async.complete();
-    });
+    httpClient.request(HttpMethod.DELETE, port, "localhost", "/_/discovery/modules")
+        .onComplete(context.asyncAssertSuccess(res -> {
+          vertx.close(x -> context.asyncAssertSuccess());
```
julianladisch, line 61: Remove `x ->` otherwise this will be called after the test has finished and will therefore not check whether vertx close has failed.

adamdickmeiss: good catch

okapi-core/src/main/java/org/folio/okapi/managers/DiscoveryManager.java
adamdickmeiss: ok

julianladisch (approved):

PR #972, opened 2020-09-14
MODAT-83 - Port of token cache PoC to Okapi 4, with metrics

craigmcnally:
See https://issues.folio.org/browse/MODAT-83
and
https://wiki.folio.org/pages/viewpage.action?pageId=36581623

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -51,6 +51,8 @@ import org.folio.okapi.common.logging.FolioLoggingContext;
 import org.folio.okapi.util.CorsHelper;
 import org.folio.okapi.util.MetricsHelper;
 import org.folio.okapi.util.ProxyContext;
+import org.folio.okapi.util.TokenCache;
+import org.folio.okapi.util.TokenCache.CacheEntry;
 
 
 /**
@@ -77,6 +79,8 @@ public class ProxyService {
   private final int waitMs;
   private static final String REDIRECTQUERY = "redirect-query"; // See redirectProxy below
   private final Messages messages = Messages.getInstance();
+
+  private TokenCache tokenCache = new TokenCache();
 
   /**
    * Construct Proxy service.
@@ -193,6 +197,8 @@ public class ProxyService {
     final String id = req.getHeader(XOkapiHeaders.MODULE_ID);
     pc.debug("getMods: Matching " + req.method() + " " + req.uri());
 
+    boolean skipAuth = false;
+
     for (ModuleDescriptor md : enabledModules) {
       pc.debug("getMods:  looking at " + md.getId());
       List<RoutingEntry> rr = null;
@@ -205,12 +211,26 @@ public class ProxyService {
         for (RoutingEntry re : rr) {
           if (match(re, req)) {
             ModuleInstance mi = new ModuleInstance(md, re, req.uri(), req.method(), true);
-            mi.setAuthToken(req.headers().get(XOkapiHeaders.TOKEN));
+
+            String pathPattern = re.getPathPattern();
+
+            // CAM
+            CacheEntry cached = tokenCache.get(pc.getTenant(), req.method().name(),
+                pathPattern == null ? req.path() : pathPattern,
+                req.headers().get(XOkapiHeaders.TOKEN), req.getHeader(XOkapiHeaders.USER_ID));
+
+            if (cached != null) {
+              mi.setAuthToken(cached.token);
+              mi.setxOkapiUserId(cached.xokapiUserid);
+              mi.setxOkapiPermissions(cached.xokapiPermissions);
+
+              skipAuth = true;
+            } else {
+              mi.setAuthToken(req.headers().get(XOkapiHeaders.TOKEN));
+            }
             mods.add(mi);
-            pc.setHandlerModuleInstance(mi);
-            pc.debug("getMods:   Added " + md.getId() + " "
-                + re.getPathPattern() + " " + re.getPath() + " "
-                + re.getPhase() + "/" + re.getLevel());
+            pc.debug("getMods:   Added " + md.getId() + " " + re.getPathPattern() + " "
+                + re.getPath() + " " + re.getPhase() + "/" + re.getLevel());
             break;
           }
         }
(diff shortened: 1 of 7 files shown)


adamdickmeiss:
Would it possible to make an Okapi issue for Okapi work? And leave the modat issue for what goes on in modat?

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -443,18 +474,37 @@ private void relayToResponse(HttpServerResponse hres,
...
           mi.setAuthToken(tok);
           pc.debug("authResponse: Default (_) token for " + id + ": " + tok);
+
+          tokenCache.put(pc.getTenant(),
+              req.method().name(),
+              pathPattern == null ? req.path() : pathPattern,
+              res.getHeader(XOkapiHeaders.USER_ID),
+              res.getHeader(XOkapiHeaders.PERMISSIONS),
+              req.getHeader(XOkapiHeaders.TOKEN),
+              tok);
```
julianladisch, line 507: duplicate code can be extracted:

    } else {
      continue;
    }
    mi.setAuthToken(tok);
    tokenCache.put(...

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -234,6 +260,11 @@ private boolean resolveRedirects(ProxyContext pc,
         -> a.getRoutingEntry().getPhaseLevel().compareTo(b.getRoutingEntry().getPhaseLevel());
     mods.sort(cmp);
 
+    if (skipAuth) {
+      pc.debug("Skipping auth, have cached token.");
+      mods.remove(0);
```
julianladisch, line 265: Do we know that the mi with skipAuth is at position 0?

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
```diff
@@ -0,0 +1,104 @@
...
+  private final Map<String, CacheEntry> cache = new HashMap<>();
+  private final long ttl;
+  
+  private static final Logger logger = OkapiLogger.get(TokenCache.class);
+  
+  public TokenCache() {
+    this(DEFAULT_TTL);
+  }
+  
+  public TokenCache(long ttl) {
```
julianladisch, line 21: Javadoc should explain what unit (minute, second, millisecond?) ttl has.

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
```diff
@@ -0,0 +1,104 @@
...
+   * @param tenant tenant id
+   * @param method HTTP method
+   * @param path path pattern
+   * @param token X-Okapi-Token header
+   * @param userId X-Okapi-User-Id header 
+   * @return cache entry or null
+   */
+  public CacheEntry get(String tenant, String method, String path, String token, String userId) {
+    String key = genKey(method, path, token, userId);
+    CacheEntry ret = cache.get(genKey(method, path, token, userId));
```
julianladisch, line 54: Should use key created in the line before.

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
```diff
@@ -0,0 +1,104 @@
...
+   */
+  public void put(String tenant, String method, String path, String userId, String xokapiPerms,
+      String keyToken, String token) {
+    long now = System.currentTimeMillis();
+    CacheEntry entry = new CacheEntry(token, userId, xokapiPerms, now + ttl);
+    String key = genKey(method, path, keyToken, userId);
+    MetricsHelper.recordTokenCacheCached(tenant, method, path, userId);
+    logger.info("Caching: {} {}", key, token);
+    cache.put(key, entry);
+  }
```
julianladisch, line 41: The cache can grow unlimited. Old entries that are expired and not used stay there forever, they are not evicted.
My suggestion:
Convert into a least recently used cache (LRUCache), examples: https://stackoverflow.com/questions/221525/how-would-you-implement-an-lru-cache-in-java
Whenever an entry is put into the cache evict the two oldest entries they are expired. This avoids a timer, is fast, and shrinks the cache if needed.

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
```diff
@@ -0,0 +1,104 @@
...
+      return null;
+    } else {
+      MetricsHelper.recordTokenCacheHit(tenant, method, path, userId);
+      logger.info("Cache Hit: {} -> {}", key, ret.token);
+      return ret;
+    }
+  }
+  
+  private String genKey(String method, String path, String token, String userId) {
+    return (method + "|" + path + "|" + token + "|" + userId).replaceAll("[\n\t\r]", "");
```
julianladisch, line 72: Can you explain why userId is added and tenantId is not added? Both are in the token.

julianladisch, line 72: It might be useful to put the userId at the beginning. This allows to easily purge the cache for a specific userId if we want to provide this feature at a later time.

craigmcnally:
Hey guys, thanks for the feedback - This is my fault for not communicating this but I never intended on merging this to master - this was the branch used for the PoC - I know it's rough and needs work before it can be merged.  I plan on creating a new branch for that (of course based off this one).  I created this PR for my own benefit - sonarqube, etc. I didn't even realize you both had reviewed this until just now.

craigmcnally:
> Would it possible to make an Okapi issue for Okapi work? And leave the modat issue for what goes on in modat?

Yes there's a separate JIRA for implementing token cache in OKAPI.  See my comment above about a separate branch for that work

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
craigmcnally: I intentionally kept this super simple for the PoC but fully intend to address these things as part of https://issues.folio.org/browse/OKAPI-890

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
craigmcnally: whoops, yup

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
craigmcnally: of course - will work this into https://issues.folio.org/browse/OKAPI-890

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
craigmcnally: I'm pretty sure it's always going to be the auth call, no?

PR #973, opened 2020-09-16
Event Bus check OKAPI-891

adamdickmeiss:
This PR makes Okapi check the Event Bus upon startup. Note that this requires that ALL Okapi instances are compiled with this check. If not, it will assume that Event Bus is not working.. Each party must be consuming on the particular message.

There's a slight chance that there's a race condition with this.. If the node list changes during startup, there might be a failure.. But this will only affect the starting Okapi process .. Not something that is already running.


okapi-core/src/main/java/org/folio/okapi/util/EventBusChecker.java
```diff
@@ -0,0 +1,51 @@
...
+public class EventBusChecker {
+  private static final String EVENT_NODE_CHECK = "org.folio.okapi.node.check.";
+
+  /**
+   * Check that event bus is operational by sending a message to each party.
+   * @param vertx Vert.x handle
+   * @param clusterManager manager ; null if not in clustered mode
+   * @return async result
+   */
+  public static Future<Void> check(Vertx vertx, ClusterManager clusterManager) {
```
julianladisch, line 20: To handle the race condition when the node list changes you may have a check3 method that runs check(Vertx, ClusterManager) up to three times if it fails.

julianladisch (approved):

hjiebsco (approved):
Curious, did we observe a case where Hazelcast formed a cluster but event bus does not work? I thought Hazelcast uses heartbeat to detect non-reachable member and updates cluster membership automatically. Anyway, the PR should help to prevent a new Okapi instance to join a non-functional cluster.

adamdickmeiss:
https://stackoverflow.com/questions/39812848/how-do-i-configure-vert-x-event-bus-to-work-across-cluster-of-docker-containers

PR #974, opened 2020-09-17
Redirect stdout/stderr to log for process deployment OKAPI-892

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
```diff
@@ -41,13 +45,14 @@
    * Construct process module handler.
    * @param vertx Vert.x handle
    * @param desc launch descriptor
+   * @param id if of process (possibly module ID)
```
julianladisch, line 48:
```suggestion
   * @param id process identifier used for logging (possibly module ID)
```

okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
```diff
@@ -98,6 +92,36 @@ public void start(Handler<AsyncResult<Void>> startFuture) {
...
+      }
+    }
+    Process process = pb.start();
+    captureStream(vertx, id, process.getInputStream());
+    captureStream(vertx, id, process.getErrorStream());
+    return process;
+  }
+
+  private static void captureStream(Vertx vertx, String id, InputStream input) {
+    vertx.executeBlocking(res -> {
```
julianladisch, line 112: The default worker pool size is 20. If I start 11 processes and each runs 2 captureStream there is some blocking. If the first runs 60 seconds and all other run 1 second then the logging of the 11th is blocked until the first has finished.

https://vertx.io/docs/vertx-core/java/#blocking_code says about executeBlocking:
> Blocking code should block for a reasonable amount of time (i.e no more than a few seconds). Long blocking operations or polling operations (i.e a thread that spin in a loop polling events in a blocking fashion) are precluded. When the blocking operation lasts more than the 10 seconds, a message will be printed on the console by the blocked thread checker. Long blocking operations should use a dedicated thread managed by the application, which can interact with verticles using the event-bus or runOnContext

See also https://github.com/vietj/childprocess-vertx-ext based on https://github.com/brettwooldridge/NuProcess :
> A low-overhead, non-blocking I/O, external Process execution implementation for Java. It is a replacement for java.lang.ProcessBuilder and java.lang.Process.
> Have you ever been annoyed by the fact that whenever you spawn a process in Java you have to create two or three "pumper" threads (for every process) to pull data out of the stdout and stderr pipes and pump data into stdin? If your code starts a lot of processes you can have dozens or hundreds of threads doing nothing but pumping data.

adamdickmeiss: Thx. I'll look for another capture approach.

okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
adamdickmeiss: Agreed. better. if of :-)

adamdickmeiss:
I really think this is good to go.  I have tried to create and destroy 100 processes (in parallel), without issues.

okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
```diff
@@ -98,6 +92,47 @@ public void start(Handler<AsyncResult<Void>> startFuture) {
     }
   }
 
+  @Override
+  public void onStdout(ByteBuffer buffer, boolean closed) {
+    byte[] bytes = new byte[buffer.remaining()];
+    buffer.get(bytes);
+    String s = new String(bytes);
```
julianladisch, line 99: This should not use the Charset of the server where Okapi runs but StandardCharsets.UTF_8.

adamdickmeiss: ok. done

julianladisch (approved):
Looks good, thanks!

PR #975, opened 2020-09-18
Okapi 872

MikhailFokanov:


okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -772,7 +772,7 @@ private static void dumpHeaders(MultiMap headers) {
     for (String name : headers.names()) {
       List<String> values = headers.getAll(name);
       for (String value : values) {
-        logger.info("{}: {}", name, value);
+        logger.debug("{}: {}", name, value);
       }
```
adamdickmeiss, line 776: We have this info dump because there was a problem with headers being too large (HTTP violation). Just to tell you why this was done..  We could set it to debug, ok. Especially since it's not a problem anymore.

adamdickmeiss (approved):
Thx

okapi-core/src/main/java/org/folio/okapi/util/ProxyContext.java
```diff
@@ -164,7 +164,7 @@ public RoutingContext getCtx() {
     return ctx;
   }
 
-  private String getReqId() {
+  private String  getReqId() {
```
hjiebsco, line 167: There is an extra space before `getReqId()`

okapi-core/src/main/java/org/folio/okapi/util/ProxyContext.java
```diff
@@ -266,23 +266,23 @@ public void addTraceHeaderLine(String h) {
...
   }
 
   public void warn(String msg) {
-    logger.warn("{} {}", getReqId(), msg);
+    logger.warn(msg);
   }
 
   public void warn(String msg, Throwable e) {
     logger.warn("{} {}", getReqId(), msg, e);
   }
```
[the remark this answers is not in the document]
MikhailFokanov: done

okapi-core/src/main/java/org/folio/okapi/util/ProxyContext.java
MikhailFokanov: done

PR #979, opened 2020-09-25
restore the missing L in Location

zburke:
Restore the missing `L` from `Location`, which doesn't actually matter
since the test for the `ocation` substring still succeeds, but this
looks nicer.

### DIFF
--- a/okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
+++ b/okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
@@ -24,9 +24,9 @@ import io.vertx.core.http.HttpServerOptions;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.streams.Pump;
+import io.vertx.ext.unit.Async;
 import io.vertx.ext.unit.TestContext;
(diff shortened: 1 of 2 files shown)


okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -1346,7 +1346,7 @@ public void testProxy(TestContext context) {
       .then().statusCode(200)
       .header("Access-Control-Allow-Origin", "*")
       .header("Access-Control-Expose-Headers", containsString(
-        "ocation,X-Okapi-Trace,X-Okapi-Token,Authorization,X-Okapi-Request-Id"))
+        "Location,X-Okapi-Trace,X-Okapi-Token,Authorization,X-Okapi-Request-Id"))
```
julianladisch, line 1349: The build fails with

```
ProxyTest.testProxy:1348 1 expectation failed.
Expected header "Access-Control-Expose-Headers" was not a string containing
"Location,X-Okapi-Trace,X-Okapi-Token,Authorization,X-Okapi-Request-Id", was 
"location,X-Okapi-Trace,X-Okapi-Token,Authorization,X-Okapi-Request-Id,X-Okapi-Module-Id".
```

julianladisch (approved):

PR #981, opened 2020-09-29
OKAPI-890 - Formalize Token Cache PoC

craigmcnally:
## Purpose
Formalize the TokenCache PoC. 

See [OKAPI-890](https://issues.folio.org/browse/OKAPI-890) and [the wiki ](https://wiki.folio.org/pages/viewpage.action?pageId=36581623) 

## Approach
* Add new config parameters:
  - `token_cache_max_size`
  - `token_cache_ttl_ms`
* Update the guide
* Implement an LRU Cache based on LinkedHashMap
* Add metrics for cache events (was part of the 2nd part of the PoC)
* Add unit tests

NOTE:  This branch is based off the [tokenCache](https://github.com/folio-org/okapi/tree/tokenCache) branch.


craigmcnally:
Do you guys think we want/need a metric (and/or even log message) when an entry is evicted from the cache?

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
```diff
@@ -0,0 +1,181 @@
...
+
+  private static final Logger logger = OkapiLogger.get(TokenCache.class);
+
+  /**
+   * Convenience constructor using the default TTL and max size.
+   */
+  public TokenCache() {
+    this(DEFAULT_TTL, DEFAULT_MAX_SIZE);
+  }
+
```
adamdickmeiss, line 25: Not in use anywhere, so I don't think it's necessary to define it.

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
```diff
@@ -0,0 +1,181 @@
...
+      logger.info("Cache Hit (Expired): {}", key);
+      cache.remove(key);
+      return null;
+    } else {
+      MetricsHelper.recordTokenCacheHit(tenant, method, path, userId);
+      logger.info("Cache Hit: {} -> {}", key, ret.token);
+      return ret;
+    }
+  }
+
```
adamdickmeiss, line 88: That's going to be a lot of logging by default.. with info log-level I mean.

craigmcnally: OK I'll make it debug

adamdickmeiss (approved):
I think the code looks good. I have not tested it in any real environment, like the Ansible box or similar.

craigmcnally:
Addressed a question that @julianladisch made in the PoC PR... instead of assuming that the first entry in the list of sorted ModuleInstances is always the auth filter (which I think it would always be anyway), it now iterates over the list and looks at the phase of each, removing the entry that has a phase matching the auth filter.  Since it breaks out of the loop when one is found, I think from a computational standpoint this will effectively be very close to just removing the first entry.

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
```diff
@@ -0,0 +1,174 @@
...
+   * @param xokapiPerms X-Okapi-Permissions header to cache
+   * @param keyToken the token to be used in the cache key, from the request which triggered the
+   *        call to mod-authtoken
+   * @param token access token to cache - from the mod-authtoken response
+   */
+  public void put(String tenant, String method, String path, String userId, String xokapiPerms,
+      String keyToken, String token) {
+    long now = System.currentTimeMillis();
+    CacheEntry entry = new CacheEntry(token, userId, xokapiPerms, now + ttl);
+    String key = genKey(method, path, userId, keyToken);
```
hjiebsco, line 47: Since `userId` is part of the key, is there a reason to store it as part of the `CacheEntry` as well?

craigmcnally: yeah I guess not.  Adam and Julian have both asked questions about the cache keys as well, I guess this needs more thought

adamdickmeiss: I think getting rid of user will be best. That means whatever content mod-authtoken puts there in the future is "safe".

okapi-core/src/main/java/org/folio/okapi/bean/ModuleInstance.java
```diff
@@ -86,4 +88,20 @@ public ModuleInstance withRetry() {
     this.withRetry = true;
     return this;
   }
+
+  public String getxOkapiUserId() {
+    return xokapiUserid;
+  }
+
+  public void setxOkapiUserId(String xokapiUserid) {
+    this.xokapiUserid = xokapiUserid;
```
julianladisch, line 97: Can you use consistent case of id, either always upper or always lower in method, parameter and variable?

okapi-core/src/main/java/org/folio/okapi/bean/ModuleInstance.java
```diff
@@ -86,4 +88,20 @@ public ModuleInstance withRetry() {
...
+  }
+
+  public String getxOkapiPermissions() {
+    return xokapiPermissions;
+  }
+
+  public void setxOkapiPermissions(String xokapiPermissions) {
+    this.xokapiPermissions = xokapiPermissions;
+  }
```
julianladisch, line 106: The lower x looks odd but it is required by Google Code Style.
Can we remove "xOkapi" from the name?

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -174,6 +187,27 @@ private boolean resolveRedirects(ProxyContext pc,
...
+        req.headers().get(XOkapiHeaders.TOKEN));
+
+    if (cached != null) {
+      mi.setAuthToken(cached.token);
+      mi.setxOkapiUserId(cached.xokapiUserid);
+      mi.setxOkapiPermissions(cached.xokapiPermissions);
+
+      skipAuth = true;
+    } else {
+      mi.setAuthToken(req.headers().get(XOkapiHeaders.TOKEN));
```
julianladisch, line 206: Can we be sure that there is no old userId and permission in mi? Or should we set them to null?

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -174,6 +187,27 @@ private boolean resolveRedirects(ProxyContext pc,
     return true;
   }
 
+  private boolean checkTokenCache(ProxyContext pc, HttpServerRequest req, RoutingEntry re,
+      ModuleInstance mi) {
+    boolean skipAuth = false;
```
julianladisch, line 192: This variable is not needed. We can directly return true or false.
There should be javadoc explaining the return value.

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
```diff
@@ -0,0 +1,174 @@
...
+      return null;
+    } else {
+      MetricsHelper.recordTokenCacheHit(tenant, method, path, userId);
+      logger.debug("Cache Hit: {} -> {}", key, ret.token);
+      return ret;
+    }
+  }
+
+  private String genKey(String method, String path, String userId, String token) {
+    return (userId + "|" + method + "|" + path + "|" + token).replaceAll("[\n\t\r]", "");
```
julianladisch, line 83: Why do we the replaceAll?
Can the replaceAll been applied to only those variables that needs it?

okapi-core/src/test/java/org/folio/okapi/util/TokenCacheTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    TokenCache cache = TokenCache.builder()
+        .withTtl(ttl)
+        .build();
+
+    cache.put("tenant", "method", "path", "userId", "xokapiPerms", "foo", "fooTok");
+
+    await().with()
+        .pollInterval(100, TimeUnit.MILLISECONDS)
+        .atMost(ttl + 100, TimeUnit.MILLISECONDS)
```
julianladisch, line 46: The test runs faster when using 50 ms for ttl and 10 ms for polling.
atMost should use ttl + 2*poll. Using ttl + poll may fail because of a race condition, we don't want spurious failures.

okapi-core/src/test/java/org/folio/okapi/util/MetricsHelperTest.java
```diff
@@ -99,6 +101,51 @@ void testRecordHttpClientResponseTime() {
     assertEquals(1, timer.count());
   }
 
+  @Test
+  void testRecordTokenCacheEvent() {
+    String userId = "03975dd7-8004-48cf-bd21-4d7ff2e74ca2";
+    String anotherUserId = "54412e3d-a024-4914-8d54-8b84e66513a6";
+
+    long ttl = 2000L;
```
julianladisch, line 109: This can be reduced to 200L

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
```diff
@@ -0,0 +1,174 @@
...
+   * @param token access token to cache - from the mod-authtoken response
+   */
+  public void put(String tenant, String method, String path, String userId, String xokapiPerms,
+      String keyToken, String token) {
+    long now = System.currentTimeMillis();
+    CacheEntry entry = new CacheEntry(token, userId, xokapiPerms, now + ttl);
+    String key = genKey(method, path, userId, keyToken);
+    MetricsHelper.recordTokenCacheCached(tenant, method, path, userId);
+    logger.debug("Caching: {} -> {}", key, token);
+    cache.put(key, entry);
```
julianladisch, line 50: The cache will grow to its max size and will never shrink because an expired token is only evicted if there is a get request. If there is a get request for an expired token then mod-authtoken is called and a new cache entry is created.

Consider to extend the put method to check the two oldest cache entries whether they are expired and to remove them.

okapi-core/src/main/java/org/folio/okapi/bean/ModuleInstance.java
craigmcnally: Good catch - typo

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
craigmcnally: I added this because sonarqube was complaining.  I suppose it could be applied to only the token (userId has been removed from the cache key in my latest changes)

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
craigmcnally: I'm not seeing any good way to get the two oldest cache entries w/o iterating the entire keySet...  Any ideas?

okapi-core/src/test/java/org/folio/okapi/util/MetricsHelperTest.java
craigmcnally: done

okapi-core/src/test/java/org/folio/okapi/util/TokenCacheTest.java
craigmcnally: done

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
craigmcnally: now only call this on token

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
craigmcnally: done

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
craigmcnally: done

okapi-core/src/main/java/org/folio/okapi/bean/ModuleInstance.java
craigmcnally: renamed them.

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
craigmcnally: Removed userId from the cache key

craigmcnally:
Note:  I [added the "iat" claim to the token in okapi-test-auth-module](https://github.com/folio-org/okapi/pull/981/commits/42616bfc5810c305f492adc1c15f52ac9837c7c3) since it was causing test failures [here](https://github.com/folio-org/okapi/blob/d5eeb07bde26ad7529b48e7f7bf46b5365696b7d/okapi-core/src/test/java/org/folio/okapi/ProxyTest.java#L3789).

The token was being cached from the first iteration (version 1.0), and that cached value being used on the second iteration, where version 1.1 is expected).  Since login is called in each iteration, adding iat to the token acts as a cache buster.  

Currently iat gets the epoch seconds (just like mod-authtoken).  In theory this might not be good enough if the two tokens are issued within the same second.  In practice it's worked fine this way for me several times in a row (albeit on a slower machine).  If this becomes an issue we can either set iat to epoch milliseconds, or inject a 1 second sleep in this loop to ensure a different token is used in this test.

adamdickmeiss (approved):
Looks good to me.

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
julianladisch: In LruCache replace K by String and V by CacheEntry. Then do this:

    @Override
    public CacheEntry put(String key, CacheEntry value) {
      CacheEntry old = super.put(key, value);

      if (size() < 3) {
        return old;
      }

      Iterator<Map.Entry<String, CacheEntry>> it = entrySet().iterator();
      for (int i=0; i<2; i++) {
        String k = it.next();
        if (expired(k)) {
          remove(k);
        } else {
          break;
        }
      }

      return old;
    }

julianladisch (approved):

okapi-core/src/main/java/org/folio/okapi/util/TokenCache.java
craigmcnally: I had to look at the javadocs again, but I now see that the iteration order is least-recently accessed to most-recently accessed, exactly what we want.  Thanks @julianladisch.

PR #982, opened 2020-10-01
K8s compatible liveness and readiness probes OKAPI-904

adamdickmeiss:
New configuration healthPort which when set and non-zero enables that Okapi
listens on that port and offers /readiness and /liveness services.. Both of
which returns 204 No Content on success. 500 on failure.


okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -1035,4 +1036,8 @@ private ModuleInstance getTenantInstanceForInterface(
     });
   }
 
+  @Override
+  public Future<Void> isAlive() {
+    return tenantStore.listTenants().mapEmpty();
```
craigmcnally, line 1041: this includes supertenant, right?

craigmcnally:
Do we plan on doing this in RMB as well?

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: yes.

okapi-core/src/test/java/org/folio/okapi/managers/HealthManagerTest.java
```diff
@@ -0,0 +1,102 @@
...
+  @Test
+  void testPortPortOK(Vertx vertx, VertxTestContext context) {
+    final int port = 9130;
+    HealthManager m = new HealthManager(port);
+    m.init(vertx, Collections.emptyList()).onComplete(context.succeedingThenComplete());
+  }
+
+  @Test
+  void testPortReadinessSuccess(Vertx vertx, VertxTestContext context) {
+    final int port = 9130;
```
julianladisch, line 34: The method should be renamed to testPortReadinessPort0 and `port = 0` been used. Then we can delete testPort0.

okapi-core/src/test/java/org/folio/okapi/managers/HealthManagerTest.java
```diff
@@ -0,0 +1,102 @@
...
+      client.get(port, "localhost", "/readiness")
+          .send(context.succeeding(response -> {
+            assertThat(response.statusCode()).isEqualTo(204);
+            context.completeNow();
+          }));
+    }));
+  }
+
+  @Test
+  void testPortReadiness(Vertx vertx, VertxTestContext context) {
```
julianladisch, line 47: This should be renamed to testPortReadinessPort9130. Then we can remove testPortOK.

okapi-core/src/test/java/org/folio/okapi/managers/HealthManagerTest.java
```diff
@@ -0,0 +1,102 @@
...
+            context.completeNow();
+          }));
+    }));
+  }
+
+  @Test
+  void testPortLivenessFailure(Vertx vertx, VertxTestContext context) {
+    final int port = 9130;
+    HealthManager m = new HealthManager(port);
+    m.init(vertx, Arrays.asList(new IsNotAlive())).onComplete(context.succeeding(res -> {
```
julianladisch, line 78: The list should be [IsAlive(), IsNotAlive()] to test that it not only checks the first one.

okapi-core/src/main/java/org/folio/okapi/managers/HealthManager.java
```diff
@@ -0,0 +1,68 @@
...
+import io.vertx.core.Vertx;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.core.http.HttpServerOptions;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+import java.util.List;
+import org.folio.okapi.service.Liveness;
+
+public class HealthManager {
+  private Vertx vertx;
```
julianladisch, line 13: This variable is not used.

okapi-core/src/main/java/org/folio/okapi/managers/HealthManager.java
```diff
@@ -0,0 +1,68 @@
...
+    future.onComplete(res -> {
+      if (res.failed()) {
+        ctx.response().putHeader("Content-Type", "text/plain");
+        ctx.response().setStatusCode(500);
+        ctx.response().end(res.cause().getMessage());
+        return;
+      }
+      ctx.response().setStatusCode(204);
+      ctx.response().end();
+    });
```
julianladisch, line 66: The idiomatic way using Future methods:
```suggestion
    future.onFailure(e -> {
      ctx.response().putHeader("Content-Type", "text/plain");
      ctx.response().setStatusCode(500);
      ctx.response().end(e.getMessage());
    }).onSuccess(handler -> {
      ctx.response().setStatusCode(204);
      ctx.response().end();
    });
```

julianladisch, line 66: Should we also log the exception (message + stacktrace) on failure? One reason might be that the database query fails. This should also be in the log, not only in the HTTP body.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
julianladisch: isAlive() succeeds if the list is empty, contains the supertenant only, or contains any number of tenants.
If only fails if the database query fails.

okapi-core/src/main/java/org/folio/okapi/managers/HealthManager.java
```diff
@@ -0,0 +1,68 @@
...
+   * @param vertx Vert.x handle
+   * @return future result
+   *
+   */
+  public Future<Void> init(Vertx vertx, List<Liveness> livenessChecks) {
+    this.vertx = vertx;
+    this.livenessChecks = livenessChecks;
+    if (listenPort == 0) {
+      return Future.succeededFuture();
+    }
```
julianladisch, line 34: The if clause should be move to the first line to avoid storing livenessChecks if they are not needed.

doc/guide.md
```diff
@@ -2737,6 +2737,8 @@ Defaults to `localhost`
 system-generated UUID (in cluster mode), or `localhost` (in dev mode)
 * `storage`: Defines the storage back end, `postgres`, `mongo` or (the default)
 `inmemory`
+* `healthPort`: port for readiness and liveness checks. Default value is
+   0 (disabled). Provide non-zero value to enable.
```
julianladisch, line 2741:
```suggestion
* `healthPort`: port for the GET /readiness and GET /liveness health checks. Use 0 to disable, this is the default. They return 204 if Okapi is ready/responsive and 500 otherwise.
```

okapi-core/src/test/java/org/folio/okapi/managers/HealthManagerTest.java
adamdickmeiss: if port = 0, then this test will start failing.. There won't be a response to /readiness, then.

okapi-core/src/test/java/org/folio/okapi/managers/HealthManagerTest.java
adamdickmeiss: ok. That will do.

okapi-core/src/test/java/org/folio/okapi/managers/HealthManagerTest.java
adamdickmeiss: Sure

okapi-core/src/main/java/org/folio/okapi/managers/HealthManager.java
adamdickmeiss: yep

okapi-core/src/main/java/org/folio/okapi/managers/HealthManager.java
adamdickmeiss: ok

okapi-core/src/main/java/org/folio/okapi/managers/HealthManager.java
adamdickmeiss: ok

doc/guide.md
adamdickmeiss: ok

julianladisch (approved):

PR #984, opened 2020-10-02
Default log4j2 logging should be patternlayout; not json OKAPI-909

adamdickmeiss:


okapi-core/src/main/resources/log4j2-json.properties
```diff
@@ -0,0 +1,38 @@
+status = error
```
julianladisch, line 1: Can you mention that Okapi supports JSON logging and add a link to log4j2-json.properties?

julianladisch: ... in guide.md?

adamdickmeiss: Good idea. Will do

julianladisch (approved):

PR #985, opened 2020-10-03
Allow module and interface discovery for same tenant OKAPI-863

adamdickmeiss:
Okapi on its own can not check if tenant is valid, so it might
pass no required permissions to auth. But auth will still fail
if tenant in token doesn't match. Internally this is done in
the routing entry of Okapi module to use a new property which
lists permissions to use when tenant matches {tenantId} in
pathPattern. This solution is, thus, backwards compatible with
OKAPI-835 for a tenant match.


okapi-core/src/test/java/org/folio/okapi/MultiTenantTest.java
```diff
@@ -335,6 +335,27 @@ public void test1() {
...
+        .header("X-Okapi-Token", okapiTokenTenant2)
+        .get("/_/proxy/tenants/" + tenant2 + "/modules")
+        .then().statusCode(200).log().ifValidationFails()
+        .extract().response();
+
+    res = given()
+        .header("Content-Type", "application/json")
+        .header("X-Okapi-Token", okapiTokenTenant2)
+        .get("/_/proxy/tenants/" + tenant1 + "/modules")
+        .then().statusCode(200).log().ifValidationFails()
```
hjiebsco, line 356: Does this mean one tenant can see another tenant's modules?

adamdickmeiss: No. In this test, the permissionsRequired are in effect and real mod-authtoken would reject because the tenant does not have permissions.

hjiebsco (approved):

PR #991, opened 2020-10-07
Fix DockerTest leaves running container OKAPI-913

adamdickmeiss:
The case where a container is started and a port is already being
in use (falure), did not stop the container as it should. This
regression was introduced with the OKAPI-899 work.


okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
```diff
@@ -363,8 +363,9 @@ private int getExposedPort(JsonObject b) {
       return createContainer(exposedPort)
           .compose(res2 -> startContainer()
               .onFailure(cause -> deleteContainer())
-              .compose(res3 -> getContainerLog())
-      );
+              .compose(res3 -> getContainerLog()
+                  .onFailure(cause -> this.stop()))
+          );
     });
```
julianladisch, line 369: Using Future methods:
```
    return getImage().map(res1 -> {
      try {
        return getExposedPort(res1);
      } catch (Exception e) {
        logger.warn("{}", e.getMessage(), e);
        throw e;
      }
    }).compose(this::createContainer)
    .compose(res2 -> startContainer().onFailure(cause -> deleteContainer()))
    .compose(res3 -> getContainerLog().onFailure(cause -> stop()));
```

`deleteContainer` should use force:

```
return deleteUrl("/containers/" + containerId + "?force=true", "deleteContainer");
```

`stop` should try to delete (with force) even if `stopContainer` has failed:

```
  @Override
  public Future<Void> stop() {
    return stopContainer()
        .compose(
            x -> deleteContainer(),
            // if stopContainer fails with e run deleteContainer but return original failure e
            e -> deleteContainer().onComplete(x -> Future.failedFuture(e)))
        .onComplete(x -> ports.free(hostPort));
  }
```

adamdickmeiss: ok. Hasn't been a problem, but it makes sense

julianladisch (approved):

PR #993, opened 2020-10-09
OKAPI-868 Add timer to capture metrics of top slow methods

hjiebsco:
PTF performance test results provided a list of slow methods in Okapi. This PR adds timer around those code blocks so the metrics can be captured. Please see https://issues.folio.org/browse/OKAPI-868.

The captured metrics in InfluxDB looks like this
```
name: org_folio_okapi_code_executionTime
time                     count host          mean      metric_type name                              sum       upper
----                     ----- ----          ----      ----------- ----                              ---       -----
2020-10-09T02:59:08.966Z 0     92.242.140.21 0         histogram   ProxyService.match                0         1.109325
2020-10-09T02:59:08.979Z 0     92.242.140.21 0         histogram   ProxyContext.logResponse          0         0.698525
2020-10-09T02:59:08.985Z 0     92.242.140.21 0         histogram   ProxyContext.logRequest           0         0.580667
2020-10-09T02:59:08.986Z 0     92.242.140.21 0         histogram   ProxyService.getModulesForRequest 0         21.003217
2020-10-09T02:59:08.986Z 3     92.242.140.21 5.290524  histogram   LockedTypedMap1.get.decodeValue   15.871572 11.978
```


okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -186,6 +189,9 @@ private boolean resolveRedirects(ProxyContext pc,
   private List<ModuleInstance> getModulesForRequest(ProxyContext pc,
                                                     List<ModuleDescriptor> enabledModules) {
```
julianladisch, line 190: This is the slowest method reported in OKAPI-868.
Can you put in more granular metrics? For example
* `ProxyService.getModulesForRequest for enabledModules`
* `ProxyService.getModulesForRequest for rr 1`
* `ProxyService.getModulesForRequest for rr 2`
* `ProxyService.getModulesForRequest sort`
* `ProxyService.getModulesForRequest for mods`

julianladisch (approved):

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
hjiebsco: Thanks @julianladisch. Looks like all slowness of `getModulesForRequest` are from `moduleManager.getEnabledModules`. I updated the [OKAPI-868](https://issues.folio.org/browse/OKAPI-868) with perf report [link](http://carrier-io.int.folio.ebsco.com/grafana/d/b3iP-8hiz/java-profiling-folio-new?orgId=1&from=1601352757170&to=1601354777273) and relevant stacktraces screenshots. Please take a look the report and let me know if you feel strongly we should add more granular metrics for `getModulesForRequest` calls.

hjiebsco: Never mind. I see what you mean. It is not about callers, but parts within `getModulesForRequest` itself.

hjiebsco: @julianladisch, I added granular metrics as suggested.

adamdickmeiss (approved):

PR #996, opened 2020-10-13
Fix unit test fails on system wo Docker OKAPI-918

adamdickmeiss:
This "issue" was introduced in OKAPI-899.


dcrossleyau (approved):
Yes. I get build success here with the command in OKAPI-918.

PR #999, opened 2020-10-14
Ignore pull image failure OKAPI-920

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
```diff
@@ -367,7 +367,7 @@ private int getExposedPort(JsonObject b) {
   @Override
   public Future<Void> start() {
     if (dockerPull) {
-      return pullImage().compose(x -> prepareContainer());
+      return pullImage().compose(x -> prepareContainer(), x -> prepareContainer());
```
hjiebsco, line 370: Maybe adding a line to explain why `x -> prepareContainer()` is called twice?

hjiebsco (approved):

okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
adamdickmeiss: It's not called twice.. Called on success and failure.. I'll use another more explicit method Future.recover.

PR #1002, opened 2020-10-16
OKAPI-921 skip token cache if no request token is present

craigmcnally:
## Overview
instead of explicitly incorporating things like tenant and userId into the cache key we use the entire token in the request as part of the cache key since it includes this information. In doing so, we need to ensure that we don't allow cache keys w/ token=null. See [OKAPI-921](https://issues.folio.org/browse/OKAPI-921)

Additional context: [FOLIO-2839](https://issues.folio.org/browse/FOLIO-2839)

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -213,8 +213,7 @@ public class ProxyService {
 
     String pathComponent = pathPattern == null ? req.path() : pathPattern;
 
-    CacheEntry cached = tokenCache == null ? null
-        : tokenCache.get(tenant, req.method().name(),
+    CacheEntry cached = tokenCache.get(tenant, req.method().name(),
             pathComponent, req.getHeader(XOkapiHeaders.USER_ID),
             token);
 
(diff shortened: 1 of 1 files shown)


adamdickmeiss:
Wonder why coverage is 0. It appers that ProxyService.java has 0% coverage on this PR.. So something is up with SQ

adamdickmeiss:
I agree that the general approach of using the entire token is desired , so that as much as possible Okapi is not coupled with what's in there.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -203,9 +203,20 @@ private boolean resolveRedirects(ProxyContext pc,
...
+      mi.setAuthToken(null);
+      mi.setUserId(null);
+      mi.setPermissions(null);
+      return false;
+    }
+
+    String pathComponent = pathPattern == null ? req.path() : pathPattern;
+
+    CacheEntry cached = tokenCache == null ? null
+        : tokenCache.get(tenant, req.method().name(),
```
adamdickmeiss, line 217: Checking for tokenCache == null seems redundant. It is always constructed. Perhaps it should be final?

adamdickmeiss (approved):

PR #1005, opened 2020-10-19
Restart modules only adds modules not yet in discovery OKAPI-922

adamdickmeiss:


hjiebsco (approved):
Just a note in case you want to improve the test coverage

adamdickmeiss:
Yes. I'm going to improve the coverage a bit. Another option for this was to just ignore the restartModules failure as we did in earlier versions.. But I think it's more honest to do it this way.

PR #1007, opened 2020-10-20
Docker pull with authenticated user OKAPI-912

adamdickmeiss:
Configuration property `dockerRegistries` includes list of registries
with authentication properties (as Docker uses them) and/or `registry` property. The latter to allow for pull on other repositories than DockerHub.

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -2752,8 +2752,9 @@ with a path like in `https://folio.example.com/okapi`.
 is. Defaults to `unix:///var/run/docker.sock`.
 * `dockerRegistries`: List of registries to use for Docker image pull. The
 value is a JSON array of objects where each object may have the following properties
-`username`, `password`, `email`, `serveraddress` and `registry`. The first 4 properties
-are passed as autentication to the Docker registry - refer to
+`username`, `password`, `email`, `serveraddress`, `identitytoken`
+and `registry`. The first 5 properties
+are passed as autentication to the Docker registry if given - refer to
 [Docker Authenticaton](https://docs.docker.com/engine/api/v1.40/#section/Authentication).
 The optional `registry` is a prefix for the image to allow pull from other
 registry than DockerHub. And empty object in the `dockerRegistries` pulls
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -2751,12 +2751,14 @@ with a path like in `https://folio.example.com/okapi`.
 * `dockerUrl`: Tells the Okapi deployment where the Docker Daemon
 is. Defaults to `unix:///var/run/docker.sock`.
 * `dockerRegistries`: List of registries to use for Docker image pull. The
-value is a JSON array of of objects where each object may have the following properties
+value is a JSON array of objects where each object may have the following properties
 `username`, `password`, `email`, `serveraddress` and `registry`. The first 4 properties
 are passed as autentication to the Docker registry - refer to
 [Docker Authenticaton](https://docs.docker.com/engine/api/v1.40/#section/Authentication).
 The optional `registry` is a prefix for the image to allow pull from other
-registry than DockerHub.
+registry than DockerHub. And empty object in the `dockerRegistries` pulls
+from DockerHub without authentication. Omitting `dockerRegistries` does
+the same and is the behavior for earlier versions of Okapi as well.
 * `containerHost`: Host where containers are running (as seen from Okapi).
 Defaults to `localhost`.
 * `postgres_host` : PostgreSQL host. Defaults to `localhost`.
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -2751,8 +2751,12 @@ with a path like in `https://folio.example.com/okapi`.
 * `dockerUrl`: Tells the Okapi deployment where the Docker Daemon
 is. Defaults to `unix:///var/run/docker.sock`.
 * `dockerRegistries`: List of registries to use for Docker image pull. The
-value is a JSON array of of objects where each object may have the following values
-for docker authentication: `username`, `password`, `email`, `serveraddress`.
+value is a JSON array of of objects where each object may have the following properties
+`username`, `password`, `email`, `serveraddress` and `registry`. The first 4 properties
+are passed as autentication to the Docker registry - refer to
+[Docker Authenticaton](https://docs.docker.com/engine/api/v1.40/#section/Authentication).
+The optional `registry` is a prefix for the image to allow pull from other
+registry than DockerHub.
 * `containerHost`: Host where containers are running (as seen from Okapi).
 Defaults to `localhost`.
 * `postgres_host` : PostgreSQL host. Defaults to `localhost`.
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -2750,6 +2750,9 @@ there happens to be one, Okapi will remove it.  Note that it may end
 with a path like in `https://folio.example.com/okapi`.
 * `dockerUrl`: Tells the Okapi deployment where the Docker Daemon
 is. Defaults to `unix:///var/run/docker.sock`.
+* `dockerRegistries`: List of registries to use for Docker image pull. The
+value is a JSON array of of objects where each object may have the following values
+for docker authentication: `username`, `password`, `email`, `serveraddress`.
 * `containerHost`: Host where containers are running (as seen from Okapi).
 Defaults to `localhost`.
 * `postgres_host` : PostgreSQL host. Defaults to `localhost`.
--- a/okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
@@ -256,7 +256,8 @@ public class DockerModuleHandle implements ModuleHandle {
         continue;
       }
       JsonObject authObject = new JsonObject();
-      for (String member : Arrays.asList("username", "password", "email", "serveraddress")) {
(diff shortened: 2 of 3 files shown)


doc/guide.md
```diff
@@ -2750,6 +2750,13 @@ there happens to be one, Okapi will remove it.  Note that it may end
 with a path like in `https://folio.example.com/okapi`.
 * `dockerUrl`: Tells the Okapi deployment where the Docker Daemon
 is. Defaults to `unix:///var/run/docker.sock`.
+* `dockerRegistries`: List of registries to use for Docker image pull. The
+value is a JSON array of of objects where each object may have the following properties
+`username`, `password`, `email`, `serveraddress` and `registry`. The first 4 properties
+are passed as autentication to the Docker registry - refer to
+[Docker Authenticaton](https://docs.docker.com/engine/api/v1.40/#section/Authentication).
```
julianladisch, line 2757: https://docs.docker.com/engine/api/v1.40/#section/Authentication allows "identitytoken", we should also support this.

doc/guide.md
```diff
@@ -2750,6 +2750,13 @@ there happens to be one, Okapi will remove it.  Note that it may end
 with a path like in `https://folio.example.com/okapi`.
 * `dockerUrl`: Tells the Okapi deployment where the Docker Daemon
 is. Defaults to `unix:///var/run/docker.sock`.
+* `dockerRegistries`: List of registries to use for Docker image pull. The
+value is a JSON array of of objects where each object may have the following properties
```
julianladisch, line 2754:
```suggestion
value is a JSON array of objects where each object may have the following properties
```

doc/guide.md
julianladisch, line 2757: Can you add an example that also shows how we support unauthenticated pulls?

okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
```diff
@@ -237,15 +242,51 @@ private void logHandler(Buffer b) {
...
-    return postUrlJson("/images/create?fromImage=" + image, "pullImage", "")
-        .mapEmpty();
+  Future<Void> pullImage() {
+    if (dockerRegistries == null) {
+      logger.info("pull image {}", image);
+      return postUrlJson("/images/create?fromImage=" + image, null, "pullImage", "")
+          .mapEmpty();
+    }
+    logger.info("pull Image using dockerRegistries");
+    Future<Boolean> future = Future.succeededFuture(Boolean.FALSE);
```
julianladisch, line 252: You can replace the Boolean by future.recover:
```
Future<Void> future = Future.failedFuture("");
for (int i = 0; i < dockerRegistries.size(); i++) {
  ...
  future = future.recover(failure -> {
    ...
    return postUrlJson("/images/create?fromImage=" + prefix + image, authObject, "pullImage", "");
  });
}
return future;
```
This has the additional benefit of returning the failure details of the last registry entry.

doc/guide.md
adamdickmeiss: I don't think so. This is returned by auth which is not used by Okapi. How long is that valid?

adamdickmeiss: That would be an empty object as one of the element for dockerRegistries .. is that the example you want? (And also the existing behavior of Okapi, when no dockerRegistries is defined).

okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
```diff
@@ -136,6 +150,110 @@ private void dockerMockHandle(RoutingContext ctx) {
...
+      Async async = context.async();
+      dh.pullImage().onComplete(context.asyncAssertFailure(x -> {
+        async.complete();
+      }));
+      async.await();
+    }
+
+    conf.put("dockerRegistries", new JsonArray()
+        .addNull()
+        .add(new JsonObject().put("username", "x").put("password", "y")));
```
julianladisch, line 201: Can you also add a test with an empty JsonObject for unauthenticated pull?
```
conf.put("dockerRegistries", new JsonArray()
        .add(new JsonObject())
        .add(new JsonObject().put("username", "x").put("password", "y")));
```

okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
adamdickmeiss: Much better. Thx

okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
adamdickmeiss: sure

doc/guide.md
adamdickmeiss: yep

doc/guide.md
adamdickmeiss: Mentioned in doc now

adamdickmeiss: Added, if somebody finds it useful :-)

okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
```diff
@@ -136,6 +150,124 @@ private void dockerMockHandle(RoutingContext ctx) {
...
+    {
+      DockerModuleHandle dh = new DockerModuleHandle(vertx, ld,
+          "mod-users-5.0.0-SNAPSHOT", ports, "localhost",
+          9231, conf);
+      Async async = context.async();
+      dh.pullImage().onComplete(context.asyncAssertSuccess(x -> {
+        async.complete();
+      }));
+      async.await();
+    }
```
julianladisch, line 185: All these code blocks are almost the same and might be extracted into a method, for example

```
boolean pullImage(TestContext context, Vertx vertx, String host, JsonObject conf) {
  LaunchDescriptor ld = new LaunchDescriptor();
  ld.setWaitIterations(2);
  ld.setDockerImage("folioci/mod-x");
  ld.setDockerPull(true);
  Ports ports = new Ports(9232, 9233);
  DockerModuleHandle dh = new DockerModuleHandle(vertx, ld,
          "mod-users-5.0.0-SNAPSHOT", ports, host,
          9231, conf);
  AtomicBoolean succeeded = new AtomicBoolean();
  dh.pullImage().onComplete(done -> {
    succeeded.set(done.succeeded()));
    async.complete();
  });
  async.await();
  return succeeded.get();
}
```
```
assertThat(pullImage(context, vertx, "localhost", conf), is(true));
```

okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
```diff
@@ -136,6 +151,68 @@ private void dockerMockHandle(RoutingContext ctx) {
...
+        .add(new JsonObject().put("username", "x").put("password", "z")));
+    context.assertTrue(pullImage(context, vertx, "localhost", conf));
+
+    conf.put("dockerRegistries", new JsonArray()
+        .add(new JsonObject().put("registry", "localhost:5000")));
+    context.assertTrue(pullImage(context, vertx, "localhost", conf));
+
+    conf.put("dockerRegistries", new JsonArray()
+        .add(new JsonObject().put("registry", "localhost:5000/")));
+    context.assertTrue(pullImage(context, vertx, "localhost", conf));
```
julianladisch, line 211: The host value was "localhost:5000" before the last commit.
If we always use "localhost" this parameter can be removed and it can be hardcoded in pullImage.
How can we unit test that the "registry" parameter is handled correctly?

okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
julianladisch: I still cannot find such a test that should succeed.

julianladisch: I still don't see it, can you point me to the line?

julianladisch (changes requested):
See the two unresolved comments.

adamdickmeiss:
Yep. It's not that easy to test the repo without one! I can fake a repo..  point my local Docker to it.. But it would hardly give a response it would like.. I could return an error and see that *that* error is propagated back. Another solution is to supply the repo of the official Docker Registry. But that might change over time.

okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
adamdickmeiss: conf.put("dockerRegistries", new JsonArray());
    context.assertFalse(pullImage(context, vertx, "localhost", conf));

doesn't succeed. Why should that be a special case?

okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
adamdickmeiss: There's one localhost:5000 and localhost:5000/ now.

adamdickmeiss: Yes. That "localhost" is the containerHost.. something completely different. And that can be hard-coded.

julianladisch (approved):

PR #1009, opened 2020-10-21
Fixes install does not honor dependencies OKAPI-925

adamdickmeiss:
But needs a good test for it.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
@@ -508,19 +508,12 @@ public class DepResolution {
   private static void upgradeLeafs(
       Map<String, ModuleDescriptor> modsAvailable,
       Map<String, ModuleDescriptor> modsEnabled, List<TenantModuleDescriptor> tml) {
-    while (true) {
-      List<String> ret = upgradeLeafs2(modsAvailable, modsEnabled, tml);
-      if (ret == null) { // nothing done
-        return;
-      }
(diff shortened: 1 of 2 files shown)


okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -506,27 +506,41 @@ private static void addOrReplace(List<TenantModuleDescriptor> tml, ModuleDescrip
...
+        return;
       }
-      ModuleDescriptor mdTo = null;
-      for (InterfaceDescriptor prov : md.getProvidesList()) {
-        for (InterfaceDescriptor req : me.getRequiresOptionalList()) {
-          if (prov.getId().equals(req.getId()) && !prov.isCompatible(req)) {
-            mdTo = lookupAvailableForProvided(modsAvailable, me, prov, mdTo);
+      // something upgraded.. try again
+    }
+  }
```
julianladisch, line 521: The list values are not needed. This can be simplified if upgradeLeafs2 returns a boolean that is true if something upgraded:

    while (upgradeLeafs2(modsAvailable, modsEnabled, tml)) {
      // something upgraded.. try again
    }

adamdickmeiss: ok

adamdickmeiss:
The List<String> ret .. is a bit misleading.. Should probably be called List<String> errors or something.. But I'd rather keep this PR not too big and refactor tests and the implementation later in master only.

julianladisch (approved):

PR #1012, opened 2020-10-22
Add more metrics backends (Prometheus and JMX) support

hjiebsco:
Added metrics support for backend Prometheus (and JMX). Documentation is updated as below. I will create a follow up ticket to move some stuff to okapi-common, so RMB can use it directly.

### Instrumentation

Okapi uses Micrometer to managing metrics and reporting to backends. To enable it,
add Java parameter `-Dvertx.metrics.options.enabled=true` first.

More Java parameters are needed to config which backends to use
* `-DinfluxDbOptions='{"uri": "http://localhost:8086", "db":"okapi"}'` - Send metrics to InfluxDB
* `-DprometheusOptions='{"embeddedServerOptions": {"port": 9930}}'` - Expose `<server>:9930/metrics` for Prometheus
* `-DjmxMetricsOptions='{"domain": "org.folio.metrics"}'` - JMX

Another Java parameter can be used to filter metrics
* `-DmetricsPrefixFilter=org.folio` - Will only report metrics with name starting `org.folio`

A full example: `java -Dvertx.metrics.options.enabled=true -DmetricsPrefixFilter=org.folio -DinfluxDbOptions='{"uri": "http://localhost:8086", "db":"okapi"}' -DprometheusOptions='{"embeddedServerOptions": {"port": 9930}}' -DjmxMetricsOptions='{"domain": "org.folio"}' -jar okapi-core/target/okapi-core-fat.jar dev`

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -3053,14 +3053,21 @@ Okapi uses Micrometer to managing metrics and reporting to backends. To enable i
 add Java parameter `-Dvertx.metrics.options.enabled=true` first.
 
 More Java parameters are needed to config which backends to use
-* `-DinfluxDbOptions='{"uri": "http://localhost:8086", "db":"okapi"}'` - Send metrics to InfluxDB
+* `-DinfluxDbOptions='{"uri": "http://localhost:8086", "db":"folio"}'` - Send metrics to InfluxDB
 * `-DprometheusOptions='{"embeddedServerOptions": {"port": 9930}}'` - Expose `<server>:9930/metrics` for Prometheus
-* `-DjmxMetricsOptions='{"domain": "org.folio.metrics"}'` - JMX
+* `-DjmxMetricsOptions='{"domain": "org.folio"}'` - JMX
 
 Another Java parameter can be used to filter metrics
 * `-DmetricsPrefixFilter=org.folio` - Will only report metrics with name starting `org.folio`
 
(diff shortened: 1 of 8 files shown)


doc/guide.md
```diff
@@ -3050,23 +3049,18 @@ Use `modulePermissions` to grant permissions for the token.
...
-* `influxUrl` - default to `http://localhost:8086`
-* `influxDbName` - default to `okapi`
-* `influxUser` - default to null
-* `influxPassword` - default to null
+Another Java parameter can be used to filter metrics
+* `-DmetricsPrefixFilter=org.folio` - Will only report metrics with name starting `org.folio`
 
-For example: `java -DinfluxUrl=http://influx.yourdomain.io:8086 -jar okapi-core/target/okapi-core-fat.jar dev -enable-metrics` then metrics
-will be sent to `http://influx.yourdomain.io:8086`
+A full example: `java -Dvertx.metrics.options.enabled=true -DmetricsPrefixFilter=org.folio -DinfluxDbOptions='{"uri": "http://localhost:8086", "db":"okapi"}' -DprometheusOptions='{"embeddedServerOptions": {"port": 9930}}' -DjmxMetricsOptions='{"domain": "org.folio"}' -jar okapi-core/target/okapi-core-fat.jar dev`
```
adamdickmeiss, line 3064: I think you should break this long line with backslashes.. Of course, so that it's possible to still cut'n paste to shell.

adamdickmeiss (approved):

doc/guide.md
hjiebsco: Reformatted as suggested.

PR #1018, opened 2020-10-26
Fix env delete does not remove from storage OKAPI-931

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/managers/EnvManager.java
```diff
@@ -86,6 +86,6 @@ public EnvManager(EnvStore s) {
   }
 
   Future<Void> remove(String name) {
-    return envMap.removeNotFound(name);
+    return envMap.removeNotFound(name).compose(res -> envStore.delete(name).mapEmpty());
```
julianladisch, line 89: There should be a unit test that fails if this fix were not applied.

adamdickmeiss: Saw this one coming :-)

adamdickmeiss: test added

julianladisch (approved):

hjiebsco (approved):

PR #1019, opened 2020-10-26
OKAPI-903: Logging object reference rather than JSON

julianladisch:
Cannot use a lambda because Mockito does not support
matching a Supplier vararg when an Object vararg exists


adamdickmeiss (approved):
ty

PR #1022, opened 2020-10-27
Cleanup ProxyContext logging OKAPI-910

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/bean/InterfaceDescriptor.java
```diff
@@ -197,11 +197,11 @@ public void setHandlers(RoutingEntry[] handlers) {
    * @param section "provides" or "requires" - the rules differ
    * @return "" if ok, or a simple error message
    */
-  public String validate(ProxyContext pc, String section, String mod) {
+  public String validate(Logger logger, String section, String mod) {
```
hjiebsco, line 200: Missing Java doc for `logger`

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -544,7 +545,7 @@ private void fireTimer(Tenant tenant, ModuleDescriptor md, RoutingEntry re, Stri
   private Future<Void> invokePermissionsForModule(Tenant tenant, ModuleDescriptor mdTo,
                                                   ModuleDescriptor permsModule, ProxyContext pc) {
 
-    pc.debug("Loading permissions for " + mdTo.getName()
+    logger.debug("Loading permissions for {}", () -> mdTo.getName()
```
hjiebsco, line 548: Would this work `logger.debug("Loading permissions for {}", mdTo.getName()`?

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -555,7 +556,7 @@ private void fireTimer(Tenant tenant, ModuleDescriptor md, RoutingEntry re, Stri
       pl = new PermissionList(moduleTo, mdTo.getExpandedPermissionSets());
     }
     String pljson = Json.encodePrettily(pl);
-    pc.debug("tenantPerms Req: " + pljson);
+    logger.debug("tenantPerms Req: {}", () -> pljson);
```
hjiebsco, line 559: Would this work `logger.debug("tenantPerms Req: {}", pljson);`?

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -572,11 +573,11 @@ private void fireTimer(Tenant tenant, ModuleDescriptor md, RoutingEntry re, Stri
           "Bad _tenantPermissions interface in module " + permsModule.getId()
               + ". No path to POST to"));
     }
-    pc.debug("tenantPerms: " + permsModule.getId() + " and " + permPath);
+    logger.debug("tenantPerms: {} and {}", (Supplier<?>) () -> permsModule.getId(), permPath);
```
hjiebsco, line 576: Would this work `logger.debug("tenantPerms: {} and {}", permsModule.getId(), permPath);`?

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -572,11 +573,11 @@ private void fireTimer(Tenant tenant, ModuleDescriptor md, RoutingEntry re, Stri
...
     return proxyService.callSystemInterface(tenant, permInst, pljson, pc).compose(cres -> {
       pc.passOkapiTraceHeaders(cres);
-      pc.debug("tenantPerms request to " + permsModule.getName()
-          + " succeeded for module " + moduleTo + " and tenant " + tenant.getId());
+      logger.debug("tenantPerms request to {} succeeded for module {} and tenant {}",
+          () -> permsModule.getId(), () -> moduleTo, () -> tenant.getId());
```
hjiebsco, line 580: Would this work `logger.debug("tenantPerms request to {} succeeded for module {} and tenant {}", permsModule.getId(), moduleTo, tenant.getId());`?

hjiebsco (approved):
Left some minor comments. One thing I noticed is that you are passing logger around for a few classes. Any advantage of doing that vs. declaring a plain static logger for each class?

okapi-core/src/main/java/org/folio/okapi/bean/InterfaceDescriptor.java
adamdickmeiss: ok

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: Yes. The question is what is fastest. In this particular case, since the getName is a fast operation, using a lambda is probably overkill.

adamdickmeiss: I'll change it to normal.. By the way.. the following line is a mess.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: yes. Since it's a simple string - probably no need to.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: yes.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: yes

adamdickmeiss:
Could made a static logger. But the nice thing about passing it is that test that it logs correctly as is done by passing a mock logger class. Could also have been done for a static logger and a setter for it.

PR #1025, opened 2020-10-29
Okapi 916 match path using hash

adamdickmeiss:
OKAPI-916... More refactoring than at first sight, because the matching.. redirects.. auth caching.. metrics was all combined in one function. The ModuleCache is "unaware" of tenants, and is managed by TenantManager so that there's one ModuleCache per tenant. The redirect functionality is ugly and not in use.. We should consider removing it.

Running through all end points for Q2 modules, the number of match invocations is down from about 500k to 5k. That's ProxyTest.testManyModules.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
@@ -398,12 +398,8 @@ public class MainVerticle extends AbstractVerticle {
   }
 
   private Future<Void> startRedeploy() {
-    return discoveryManager.restartModules().compose(res -> {
-      if (!enableProxy) {
-        return Future.succeededFuture();
-      }
-      return tenantManager.startTimers(discoveryManager);
-    });
+    return discoveryManager.restartModules()
+        .compose(res -> tenantManager.startTimers(discoveryManager));
   }
 
 }
--- a/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
@@ -163,7 +163,6 @@ public class MainVerticle extends AbstractVerticle {
       moduleManager = new ModuleManager(moduleStore);
       TenantStore tenantStore = storage.getTenantStore();
       tenantManager = new TenantManager(moduleManager, tenantStore);
-      moduleManager.setTenantManager(tenantManager);
       discoveryManager.setModuleManager(moduleManager);
       logger.info("Proxy using {} storage", storageType);
       PullManager pullManager = new PullManager(vertx, moduleManager);
@@ -179,7 +178,6 @@ public class MainVerticle extends AbstractVerticle {
       moduleManager.forceLocalMap(); // make sure it is not shared
       tenantManager = new TenantManager(moduleManager, new TenantStoreNull());
       tenantManager.forceLocalMap();
-      moduleManager.setTenantManager(tenantManager);
       discoveryManager.setModuleManager(moduleManager);
       InternalModule internalModule = new InternalModule(
           null, null, deploymentManager, null,
--- a/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
@@ -265,7 +265,7 @@ public class MainVerticle extends AbstractVerticle {
           okapiModule, interfaceVersion);
       moduleManager.create(md, true, true, true).onFailure(cause1 ->
           promise.fail(cause1) // something went badly wrong
-      ).onComplete(ires -> {
+      ).onSuccess(ires -> {
         checkSuperTenant(okapiModule, promise);
       });
     });
--- a/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
@@ -44,7 +44,6 @@ import org.folio.okapi.service.impl.TenantStoreNull;
 import org.folio.okapi.util.CorsHelper;
 import org.folio.okapi.util.EventBusChecker;
 import org.folio.okapi.util.LogHelper;
-import org.folio.okapi.util.MetricsHelper;
 import org.folio.okapi.util.OkapiError;
 
 @java.lang.SuppressWarnings({"squid:S1192"})
--- a/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
@@ -250,25 +250,22 @@ public class MainVerticle extends AbstractVerticle {
     final ModuleDescriptor md = InternalModule.moduleDescriptor(okapiVersion);
     final String okapiModule = md.getId();
     final String interfaceVersion = md.getProvides()[0].getVersion();
-    moduleManager.get(okapiModule).onComplete(gres -> {
-      if (gres.succeeded()) { // we already have one, go on
-        logger.debug("checkInternalModules: Already have {} "
-            + " with interface version {}", okapiModule, interfaceVersion);
-        // See Okapi-359 about version checks across the cluster
-        checkSuperTenant(okapiModule, promise);
-        return;
-      }
-      if (OkapiError.getType(gres.cause()) != ErrorType.NOT_FOUND) {
-        promise.fail(gres.cause()); // something went badly wrong
+    moduleManager.get(okapiModule).onSuccess(gres -> {
+      // we already have one, go on
+      logger.debug("checkInternalModules: Already have {} "
+          + " with interface version {}", okapiModule, interfaceVersion);
+      // See Okapi-359 about version checks across the cluster
+      checkSuperTenant(okapiModule, promise);
+    }).onFailure(cause -> {
+      if (OkapiError.getType(cause) != ErrorType.NOT_FOUND) {
+        promise.fail(cause); // something went badly wrong
         return;
       }
       logger.debug("Creating the internal Okapi module {} with interface version {}",
           okapiModule, interfaceVersion);
-      moduleManager.create(md, true, true, true).onComplete(ires -> {
-        if (ires.failed()) {
-          promise.fail(ires.cause()); // something went badly wrong
-          return;
-        }
+      moduleManager.create(md, true, true, true).onFailure(cause1 ->
+          promise.fail(cause1) // something went badly wrong
+      ).onComplete(ires -> {
         checkSuperTenant(okapiModule, promise);
       });
     });
@@ -276,52 +273,48 @@ public class MainVerticle extends AbstractVerticle {
   }
 
   private void checkSuperTenant(String okapiModule, Promise<Void> promise) {
-    tenantManager.get(XOkapiHeaders.SUPERTENANT_ID).onComplete(gres -> {
-      if (gres.succeeded()) { // we already have one, go on
-        logger.info("checkSuperTenant: Already have " + XOkapiHeaders.SUPERTENANT_ID);
-        Tenant st = gres.result();
-        Set<String> enabledMods = st.getEnabled().keySet();
-        if (enabledMods.contains(okapiModule)) {
-          logger.info("checkSuperTenant: enabled version is {}", okapiModule);
-          promise.complete();
-          return;
-        }
-        // Check version compatibility
-        String enver = "";
-        for (String emod : enabledMods) {
-          if (emod.startsWith("okapi-")) {
-            enver = emod;
-          }
-        }
-        final String ev = enver;
-        logger.debug("checkSuperTenant: Enabled version is '{}', not '{}'",
-            ev, okapiModule);
-        // See Okapi-359 about version checks across the cluster
-        if (ModuleId.compare(ev, okapiModule) >= 4) {
-          logger.warn("checkSuperTenant: This Okapi is too old,"
-                  + "{} we already have {} in the database. Use that!",
-              okapiVersion, ev);
-          promise.complete();
-          return;
-        }
-        logger.info("checkSuperTenant: Need to upgrade the stored version from {} to {}",
-            ev, okapiModule);
-        // Use the commit, easier interface.
-        // the internal module can not have dependencies
-        // See Okapi-359 about version checks across the cluster
-        tenantManager.updateModuleCommit(st, ev, okapiModule).onComplete(ures -> {
-          if (ures.failed()) {
-            promise.fail(ures.cause());
-            return;
-          }
-          logger.info("Upgraded the InternalModule version from '{}' to '{}' for {}",
-              ev, okapiModule, XOkapiHeaders.SUPERTENANT_ID);
-          promise.complete();
-        });
+    tenantManager.get(XOkapiHeaders.SUPERTENANT_ID).onSuccess(tenant -> {
+      // we already have one, go on
+      logger.info("checkSuperTenant: Already have " + XOkapiHeaders.SUPERTENANT_ID);
+      Set<String> enabledMods = tenant.getEnabled().keySet();
+      if (enabledMods.contains(okapiModule)) {
+        logger.info("checkSuperTenant: enabled version is {}", okapiModule);
+        promise.complete();
         return;
       }
-      if (OkapiError.getType(gres.cause()) != ErrorType.NOT_FOUND) {
-        promise.fail(gres.cause()); // something went badly wrong
+      // Check version compatibility
+      String enver = "";
+      for (String emod : enabledMods) {
+        if (emod.startsWith("okapi-")) {
+          enver = emod;
+        }
+      }
+      final String ev = enver;
+      logger.debug("checkSuperTenant: Enabled version is '{}', not '{}'",
+          ev, okapiModule);
+      // See Okapi-359 about version checks across the cluster
+      if (ModuleId.compare(ev, okapiModule) >= 4) {
+        logger.warn("checkSuperTenant: This Okapi is too old,"
+                + "{} we already have {} in the database. Use that!",
+            okapiVersion, ev);
+        promise.complete();
+        return;
+      }
+      logger.info("checkSuperTenant: Need to upgrade the stored version from {} to {}",
+          ev, okapiModule);
+      // Use the commit, easier interface.
+      // the internal module can not have dependencies
+      // See Okapi-359 about version checks across the cluster
+      tenantManager.updateModuleCommit(tenant, ev, okapiModule).onFailure(cause ->
+          promise.fail(cause)
+      ).onSuccess(res -> {
+        logger.info("Upgraded the InternalModule version from '{}' to '{}' for {}",
+            ev, okapiModule, XOkapiHeaders.SUPERTENANT_ID);
+        promise.complete();
+      });
+    }).onFailure(cause -> {
+      if (OkapiError.getType(cause) != ErrorType.NOT_FOUND) {
+        promise.fail(cause); // something went badly wrong
         return;
       }
       logger.info("Creating the superTenant " + XOkapiHeaders.SUPERTENANT_ID);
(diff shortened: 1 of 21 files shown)


okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -1044,6 +1085,68 @@ private ModuleInstance getTenantInstanceForInterface(
...
+   * Get module cache for tenant.
+   * @param tenant Tenant
+   * @return Module Cache
+   */
+  public Future<ModuleCache> getModuleCache(Tenant tenant) {
+    if (!enabledModulesCache.containsKey(tenant.getId())) {
+      return Future.succeededFuture(new ModuleCache(new LinkedList<>()));
+    }
+    return Future.succeededFuture(enabledModulesCache.get(tenant.getId()));
+  }
```
julianladisch, line 1098: This should be replaced by 

`return Future.succeededFuture(enabledModulesCache.computeIfAbsent(tenant.getId(), key -> new ModuleCache(new LinkedList<>())));`

adamdickmeiss: More elegant indeed.

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
```diff
@@ -904,7 +904,14 @@ public static ModuleDescriptor moduleDescriptor(String okapiVersion) {
...
 
   private Future<String> deleteModule(String id) {
-    return moduleManager.delete(id).compose(res -> Future.succeededFuture(""));
+    return tenantManager.getModuleUser(id)
+        .compose(tenants -> {
+          if (!tenants.isEmpty()) {
+            return Future.failedFuture(new OkapiError(ErrorType.USER,
+                messages.getMessage("10206", id, tenants.get(0))));
+          }
+          return moduleManager.delete(id).compose(res -> Future.succeededFuture(""));
```
julianladisch, line 913:
```suggestion
          return moduleManager.delete(id).map("");
```

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -664,16 +560,16 @@ public void proxy(RoutingContext ctx) {
         pc.responseError(400, messages.getMessage("10106", tenantId));
         return;
       }
-      Tenant tenant = gres.result();
-      moduleManager.getEnabledModules(tenant).onComplete(mres -> {
-        if (mres.failed()) {
+      tenantManager.getModuleCache(gres.result()).onComplete(cacheRes -> {
+        if (cacheRes.failed()) {
```
julianladisch, line 564:
```
.onComplete(cacheRes -> {
  if (cacheRes.failed()) {
    FAILUREBLOCK
    return;
  }
  ModuleCache cache = cacheRes.result();
  SUCCESSBLOCK
});
```
should be replaced by
```
.onFailure(cause -> {
  FAILUREBLOCK
}).onSuccess(cache -> {
  SUCCESSBLOCK
});
```

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
adamdickmeiss: ok

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: ok

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
```diff
@@ -625,10 +625,22 @@ public void testFilters(TestContext context) {
 
     // Clean up (in reverse order)
     logger.debug("testFilters starting to clean up");
-    given().delete(locPostEnable).then().log().ifValidationFails().statusCode(204);
-    given().delete(locationPostDeployment).then().log().ifValidationFails().statusCode(204);
+    c = api.createRestAssured3();
+    c.given().delete(locPostEnable).then().log().ifValidationFails().statusCode(204);
+    Assert.assertTrue("raml: " + c.getLastReport().toString(),
+        c.getLastReport().isEmpty());
```
julianladisch, line 631: Can you extract this duplicated code into some method, for example `assertEmptyReport(RestAssuredClient c)`?

okapi-core/src/test/java/org/folio/okapi/util/ModuleCacheTest.java
```diff
@@ -0,0 +1,350 @@
...
+      "/a/b, /a/b",
+      "/a/{id}, /a/",
+  })
+  void testGetPatternPrefix(ArgumentsAccessor accessor) {
+    RoutingEntry routingEntry = new RoutingEntry();
+    routingEntry.setPathPattern(accessor.getString(0));
+    String expect = accessor.getString(1);
+    if (expect == null) {
+      expect = "";
+    }
```
julianladisch, line 39: You can directly provide an empty string when you use CsvSource' quoting:
`"{id}, ''"`

okapi-core/src/test/java/org/folio/okapi/util/ModuleCacheTest.java
```diff
@@ -0,0 +1,350 @@
...
+      "/{id}, /",
+      "{id},",
+      "/a/b{id}, /a/",
+      "/a/*c, /a/",
+      "/a/b*, /a/",
+      "/a/b/, /a/b/",
+      "/a/b, /a/b",
+      "/a/{id}, /a/",
+  })
+  void testGetPatternPrefix(ArgumentsAccessor accessor) {
```
julianladisch, line 33:
```suggestion
  void testGetPatternPrefix(String pathPattern, String expect) {
```

okapi-core/src/test/java/org/folio/okapi/util/ModuleCacheTest.java
adamdickmeiss: ok

okapi-core/src/test/java/org/folio/okapi/util/ModuleCacheTest.java
adamdickmeiss: ok

okapi-core/src/test/java/org/folio/okapi/util/ModuleCacheTest.java
```diff
@@ -0,0 +1,350 @@
...
+
+    Map<String, List<ModuleCache.ModuleCacheEntry>> map = new HashMap<>();
+    ModuleCache.add(md, map, routingEntries);
+
+    assertThat(ModuleCache.lookup("", HttpMethod.GET, map, true, null)).isEmpty();
+    assertThat(ModuleCache.lookup("/", HttpMethod.GET, map, true, null)).isEmpty();
+    assertThat(ModuleCache.lookup("/a", HttpMethod.GET, map, true, null)).isEmpty();
+    List<ModuleInstance> instances = ModuleCache.lookup("/a/b", HttpMethod.GET, map, true, null);
+    assertThat(instances.size()).isEqualTo(1);
+    assertThat(instances.get(0).getRoutingEntry()).isEqualTo(routingEntry1);
```
julianladisch, line 99: http://hamcrest.org/JavaHamcrest/javadoc/2.2/org/hamcrest/Matchers.html#contains-E...-
allows to simply use

    assertThat(instances, contains(routingEntry1));

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
```diff
@@ -253,78 +250,71 @@ public void start(Promise<Void> promise) {
...
       logger.debug("Creating the internal Okapi module {} with interface version {}",
           okapiModule, interfaceVersion);
-      moduleManager.create(md, true, true, true).onComplete(ires -> {
-        if (ires.failed()) {
-          promise.fail(ires.cause()); // something went badly wrong
-          return;
-        }
+      moduleManager.create(md, true, true, true).onFailure(cause1 ->
+          promise.fail(cause1) // something went badly wrong
+      ).onComplete(ires -> {
```
hjiebsco, line 268: Is this supposed to be `onSuccess` or `onComplete`? Do you want `checkSuperTenant` always be done?

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
adamdickmeiss: ok (in use in other tests, though, so could also move it to a test util).

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
adamdickmeiss: It should be onSuccess  (and that was the existing behavior).

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -132,62 +133,13 @@ private void makeTraceHeader(ModuleInstance mi, int statusCode,
     pc.logResponse(mi.getModuleDescriptor().getId(), url, statusCode);
   }
 
-  private boolean match(RoutingEntry e, HttpServerRequest req) {
+  private static boolean match(RoutingEntry e, HttpServerRequest req) {
```
hjiebsco, line 136: Looks like this method is not used anymore.

adamdickmeiss: Right

okapi-core/src/test/java/org/folio/okapi/util/ModuleCacheTest.java
```diff
@@ -0,0 +1,350 @@
...
+
+    instances = ModuleCache.lookup("/a/b", HttpMethod.GET, map, true, "module-1.0.0");
+    assertThat(instances.size()).isEqualTo(1);
+    assertThat(instances.get(0).getRoutingEntry()).isEqualTo(routingEntry1);
+
+    assertThat(ModuleCache.lookup("/a", HttpMethod.GET, map, true, "other-1.0.0")).isEmpty();
+
+    instances = ModuleCache.lookup("/a/b", HttpMethod.POST, map, true, null);
+    assertThat(instances.size()).isEqualTo(1);
+    assertThat(instances.get(0).getRoutingEntry()).isEqualTo(routingEntry2);
```
julianladisch, line 105: If you have

    Function<ModuleInstance, RoutingEntry> routingEntry = ModuleInstance::getRoutingEntry;

you can reduce the two lines to a one-liner:

    assertThat(instances).extracting(routingEntry).containsExactly(routingEntry1);

adamdickmeiss: ok.. I can tell you.. You'll be able to find many constructs  - for example in ModuleTest - that could be simplified.

okapi-core/src/test/java/org/folio/okapi/util/ModuleCacheTest.java
```diff
@@ -0,0 +1,346 @@
...
+    assertThat(instances.size()).isEqualTo(1);
+    assertThat(instances.get(0).getRoutingEntry()).isEqualTo(routingEntry3);
+
+    instances = ModuleCache.lookup("/p/id/y", HttpMethod.GET, map, true, null);
+    assertThat(instances.size()).isEqualTo(1);
+    assertThat(instances.get(0).getRoutingEntry()).isEqualTo(routingEntry4);
+
+    ModuleCache.lookup("/p/id/z", HttpMethod.GET, map, true, null).isEmpty();
+    ModuleCache.lookup("/p/id", HttpMethod.GET, map, true, null).isEmpty();
+    ModuleCache.lookup("/p/id/y/z", HttpMethod.GET, map, true, null).isEmpty();
```
julianladisch, line 120: assertThat is missing

okapi-core/src/test/java/org/folio/okapi/util/ModuleCacheTest.java
```diff
@@ -0,0 +1,346 @@
...
+
+    instances = ModuleCache.lookup("/perms/users/y", HttpMethod.GET, map, true, null);
+    assertThat(instances.size()).isEqualTo(1);
+    assertThat(instances.get(0).getRoutingEntry()).isEqualTo(routingEntry5);
+
+    instances = ModuleCache.lookup("/perms/users1", HttpMethod.GET, map, true, null);
+    assertThat(instances.size()).isEqualTo(1);
+    assertThat(instances.get(0).getRoutingEntry()).isEqualTo(routingEntry5);
+
+    ModuleCache.lookup("/perms/user", HttpMethod.GET, map, true, null).isEmpty();
```
julianladisch, line 134: assertThat is missing

adamdickmeiss: good catch

okapi-core/src/test/java/org/folio/okapi/util/ModuleCacheTest.java
adamdickmeiss: yep. Corrected.

okapi-core/src/test/java/org/folio/okapi/util/ModuleCacheTest.java
julianladisch: Thanks, more readable now.

okapi-core/src/main/java/org/folio/okapi/util/ModuleCache.java
```diff
@@ -0,0 +1,203 @@
...
+    while (index < pathPattern.length()) {
+      if (pathPattern.charAt(index) == '*' || pathPattern.charAt(index) == '{') {
+        while (index > 0 && pathPattern.charAt(index - 1) != '/') {
+          --index;
+        }
+        break;
+      }
+      index++;
+    }
+    return pathPattern.substring(0, index);
```
julianladisch, line 71:
```suggestion
    int lastSlash = 0;
    for (int i = 0; i < pathPattern.length(); i++) {
      switch (pathPattern.charAt(i)) {
        case '*':
        case '{':
          return pathPattern.substring(0, lastSlash);
        case '/':
          lastSlash = i + 1;
          break;
        default:
          break;
      }
    }
    return pathPattern;
```

okapi-core/src/main/java/org/folio/okapi/util/ModuleCache.java
```diff
@@ -0,0 +1,203 @@
...
+        }
+      }
+      int index = tryUri.length() - 1;
+      while (index > 0 && tryUri.charAt(index - 1) != '/') {
+        --index;
+      }
+      if (index <= 0) {
+        break;
+      }
+      tryUri = tryUri.substring(0, index);
```
julianladisch, line 166: We can use lastIndexOf:
```suggestion
      int index = tryUri.lastIndexOf('/', tryUri.length() - 2);
      if (index < 0) {
        break;
      }
      tryUri = tryUri.substring(0, index + 1);
```

julianladisch (approved):
Very nice, thanks!

okapi-core/src/main/java/org/folio/okapi/util/ModuleCache.java
adamdickmeiss: Ok. A little more lines, but perhaps more readable. In particular it avoids substring when there's no masking.

okapi-core/src/main/java/org/folio/okapi/util/ModuleCache.java
adamdickmeiss: Definitely shorter.

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
```diff
@@ -253,78 +250,71 @@ public void start(Promise<Void> promise) {
...
+      }
+      final String ev = enver;
+      logger.debug("checkSuperTenant: Enabled version is '{}', not '{}'",
+          ev, okapiModule);
+      // See Okapi-359 about version checks across the cluster
+      if (ModuleId.compare(ev, okapiModule) >= 4) {
+        logger.warn("checkSuperTenant: This Okapi is too old,"
+                + "{} we already have {} in the database. Use that!",
+            okapiVersion, ev);
+        promise.complete();
```
hjiebsco, line 300: This PR did not change this part, but I am wondering if we should consider downgrade Okapi version for superteant so it matches the actual running version?

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -228,89 +173,36 @@ private boolean checkTokenCache(String tenant, HttpServerRequest req, String pat
...
-            MetricsHelper.recordCodeExecutionTime(sampleLoopEnabledModules,
-                name + ".loopEnabledModules");
-            MetricsHelper.recordCodeExecutionTime(sample, name);
-            return null;
-          }
-          logger.debug("getModulesForRequest:  Added {} {} {} {} / {}", md.getId(),
-              re.getPathPattern(), re.getPath(), re.getPhase(), re.getLevel());
-        }
+    }
+    for (ModuleInstance mi : mods) {
```
hjiebsco, line 196: Can this loop be combined with above one?

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -228,89 +173,36 @@ private boolean checkTokenCache(String tenant, HttpServerRequest req, String pat
...
       }
-      MetricsHelper.recordCodeExecutionTime(sampleLoopFilterEntries, name + ".loopFilterEntries");
     }
-    MetricsHelper.recordCodeExecutionTime(sampleLoopEnabledModules, name + ".loopEnabledModules");
-
-    Timer.Sample sampleSortModuleInstances = MetricsHelper.getTimerSample();
-    Comparator<ModuleInstance> cmp = (ModuleInstance a, ModuleInstance b)
-        -> a.getRoutingEntry().getPhaseLevel().compareTo(b.getRoutingEntry().getPhaseLevel());
+    Comparator<ModuleInstance> cmp = Comparator.comparing((ModuleInstance a)
+        -> a.getRoutingEntry().getPhaseLevel());
```
hjiebsco, line 202: Wondering if we should extract this so it won't be executed every time?

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
adamdickmeiss: We could it it's just the Okapi module .. But we do have to be careful.. Eg ModuleDescriptors with new properties that suddenly will not be recognized by Okapi anymore.. I'll see what I can do as part of the "upgrade Okapi module for all tenants".. https://issues.folio.org/browse/OKAPI-930

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: Yes. Will do.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: I don't see how we could save this sorting.. Having said that we're sorting - in most cases two modules: auth + handler module.. Suppose a few audit.. and we're up to 4.

hjiebsco: I mean define it just once somewhere and then use it every time.

okapi-core/src/main/java/org/folio/okapi/util/ModuleCache.java
```diff
@@ -0,0 +1,203 @@
...
+   * @param method HTTP method
+   * @param id Proxy-ID for multi lookup; otherwise null
+   * @return module instances that match
+   * @throws IllegalArgumentException for redirect errors
+   */
+  public List<ModuleInstance> lookup(String uri, HttpMethod method, String id) {
+    logger.debug("lookup {} {} id={}", () -> method.name(), () -> uri, () -> id);
+    logger.debug("Available modules {}", () -> ModuleUtil.moduleList(moduleDescriptors));
+    // perform lookup of filters
+    List<ModuleInstance> instances = ModuleCache.lookup(uri, method, filterMap, false, null);
```
hjiebsco, line 183: Can remove `ModuleCache` prefix since it is in the same class.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: ok. done

hjiebsco (approved):

PR #1032, opened 2020-11-04
OKAPI-943: GenericCompositeFuture

julianladisch:
Coordinate a List<Future<T extends Future <?>>>, provides
all(List<T>), any(List<T>) and join(List<T>) that work like
CompositeFuture#all(List), CompositeFuture#any(List) and
CompositeFuture#join(List) but properly check the
generic type <T extends Future<?>>.

See https://github.com/eclipse-vertx/vert.x/pull/3595 why Vert.x doesn't have this.

### DIFF
--- a/okapi-common/pom.xml
+++ b/okapi-common/pom.xml
@@ -71,6 +71,11 @@
       <artifactId>vertx-unit</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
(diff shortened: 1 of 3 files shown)


adamdickmeiss:
WARNING] src/main/java/org/folio/okapi/common/GenericCompositeFuture.java:[15] (javadoc) JavadocParagraph: <p> tag should be placed immediately before the first word, with no space after.

[WARNING] src/main/java/org/folio/okapi/common/GenericCompositeFuture.java:[15] (javadoc) JavadocParagraph: <p> tag should be preceded with an empty line.

adamdickmeiss (approved):
Yes. It's really been annoying with the existing implementation.

PR #1035, opened 2020-11-05
Link Docker and Ubuntu repositories

julianladisch:
To qualify for the Open Source Project status under the
`Docker Open Source Policy` FOLIO as a publisher needs to
comply with the `Joint Marketing Programs` requirement:

"While the publisher retains the Open Source project status, the Publisher agrees to -
* Become a Docker public reference for press releases, blogs, webinars, etc
* Create joint blogs, webinars and other marketing content
* Create explicit links to their Docker Hub repos, with no ‘wrapping’ or hiding sources of their images
* Document that Docker Engine or Docker Desktop are required to run their whitelisted images
* Give Docker full attribution"

Okapi is the first pinned repository on https://github.com/folio-org

The Docker text has been copied from
https://dev.folio.org/download/artifacts/

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,28 @@ For example:
 When filing bug reports with unit test output, always use Maven
 option `-B` to avoid control characters in output.
 
+## Docker image
+
+At Docker Hub:
+
+* https://hub.docker.com/u/folioorg/okapi released versions
+* https://hub.docker.com/r/folioci/okapi snapshot versions
+
+See [Automation/Docker
+Hub](https://dev.folio.org/guides/automation/#docker-hub) for details.
+
+Docker images are the primary distribution model for FOLIO modules.
+To run the images you will need the Docker Engine or Docker Desktop
+runtime.
+
+## Ubuntu package
+
+Import the FOLIO signing key and add the [FOLIO apt
+repository](https://repository.folio.org/packages/ubuntu/):
+
+    wget --quiet -O - https://repository.folio.org/packages/debian/folio-apt-archive-key.asc | sudo apt-key add -
+    sudo add-apt-repository "deb https://repository.folio.org/packages/ubuntu/ focal/"
+
 ## Documentation
 
 * [Okapi Guide and Reference](doc/guide.md)
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ option `-B` to avoid control characters in output.
 
 At Docker Hub:
 
-* https://hub.docker.com/u/folioorg/okapi released versions
+* https://hub.docker.com/r/folioorg/okapi released versions
 * https://hub.docker.com/r/folioci/okapi snapshot versions
 
 See [Automation/Docker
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,6 +42,28 @@ For example:
 When filing bug reports with unit test output, always use Maven
 option `-B` to avoid control characters in output.
 
+## Docker image
+
+At Docker Hub:
+
+* https://hub.docker.com/u/folioorg/okapi released versions
```
hjiebsco, line 49: Looks like the correct link should be `https://hub.docker.com/r/folioorg/okapi`. @dcrossleyau, can you confirm? If yes, we need to update https://dev.folio.org/download/artifacts/ as well.

julianladisch: Yes, good catch. I should have copied the complete URL, not appending `/okapi` to the `https://hub.docker.com/u/folioorg` URL.
Fixed.

README.md
```diff
@@ -42,6 +42,28 @@ For example:
...
+Docker images are the primary distribution model for FOLIO modules.
+To run the images you will need the Docker Engine or Docker Desktop
+runtime.
+
+## Ubuntu package
+
+Import the FOLIO signing key and add the [FOLIO apt
+repository](https://repository.folio.org/packages/ubuntu/):
+
+    wget --quiet -O - https://repository.folio.org/packages/debian/folio-apt-archive-key.asc | sudo apt-key add -
```
adamdickmeiss, line 64: Perhaps use -q instead of --quiet ?

README.md
```diff
@@ -42,6 +42,28 @@ For example:
...
+See [Automation/Docker
+Hub](https://dev.folio.org/guides/automation/#docker-hub) for details.
+
+Docker images are the primary distribution model for FOLIO modules.
+To run the images you will need the Docker Engine or Docker Desktop
+runtime.
+
+## Ubuntu package
+
+Import the FOLIO signing key and add the [FOLIO apt
```
adamdickmeiss, line 61: This is not for any Ubuntu.. Just Ubuntu focal 20.04 LTS.. So perhaps header should be changed or there could be a leading paragraph saying that.

adamdickmeiss (approved):

README.md
julianladisch: If prefer not to mention the Ubuntu release in the header so that links to this section will not break.

Clicking on https://repository.folio.org/packages/ubuntu/ shows that we have packages for xenial and focal. However, latest packages are available for focal only.

I've added "Ubuntu 20.04 LTS (Focal Fossa)" to the text.

README.md
julianladisch: Replaced.

adamdickmeiss (approved):

dcrossleyau (approved):

README.md
dcrossleyau: I fixed the namespace URLs at dev.f.o

hjiebsco (approved):

PR #1036, opened 2020-11-06
Upgrade okapi module for tenants OKAPI-930

adamdickmeiss:
Updates okapi modules for all tenants that have okapi module enabled already (supertenant always has okapi module enabled by default). Downgrade also possible, but it will refuse to downgrade to major lower version.

This PR also include a serious error for the mongo storage.

Also includes various fixes to MainDeployTest plus major clean up of ModuleTest.


okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
```diff
@@ -244,93 +243,45 @@ public void start(Promise<Void> promise) {
...
           okapiModule, interfaceVersion);
-      moduleManager.create(md, true, true, true).onFailure(cause1 ->
-          promise.fail(cause1) // something went badly wrong
-      ).onSuccess(ires -> {
-        checkSuperTenant(okapiModule, promise);
-      });
-    });
-    return promise.future();
+      return moduleManager.create(md, true, true, true);
+    }).compose(x -> checkSuperTenant(okapiModule));
```
julianladisch, line 265: This indentation allows to remove `@SuppressWarnings("indentation")`:

    return moduleManager.get(okapiModule).compose(
        gres -> {
          // we already have one, go on
          logger.debug("checkInternalModules: Already have {} "
              + " with interface version {}", okapiModule, interfaceVersion);
          // See Okapi-359 about version checks across the cluster
          return Future.succeededFuture();
        },
        cause -> {
          if (OkapiError.getType(cause) != ErrorType.NOT_FOUND) {
            return Future.failedFuture(cause); // something went badly wrong
          }
          logger.debug("Creating the internal Okapi module {} with interface version {}",
              okapiModule, interfaceVersion);
          return moduleManager.create(md, true, true, true);
        })
        .compose(x -> checkSuperTenant(okapiModule));

adamdickmeiss: Nice. Thx

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -193,15 +185,12 @@ private void invalidateCacheEntry(String id) {
   }
 
   /**
-   * Get a module descriptor from ID.
+   * Get a module descriptor from ID and return Error if not found.
```
julianladisch, line 188:
```suggestion
   * Get a module descriptor from ID, if not found the Future will fail with
   * an {@link OkapiError} with {@link ErrorType#NOT_FOUND}.
```

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -268,30 +247,30 @@ public void setProxyService(ProxyService px) {
...
+    return future1.compose(mdTo -> {
+      Future<ModuleDescriptor> future2 = moduleFrom != null
+          ? moduleManager.get(moduleFrom) : Future.succeededFuture(null);
+      return future2.compose(mdFrom -> {
+        Future<Void> future3 = options.getDepCheck()
+            ? enableAndDisableCheck(tenant, mdFrom, mdTo) : Future.succeededFuture();
+        return future3
+            .compose(x -> enableAndDisableModule(tenant, options, mdFrom, mdTo, pc));
+      });
+    });
```
julianladisch, line 273:
```suggestion
    Future<ModuleDescriptor> mdFrom = moduleFrom != null
        ? moduleManager.get(moduleFrom) : Future.succeededFuture(null);
    Future<ModuleDescriptor> mdTo = moduleTo != null
        ? moduleManager.getLatest(moduleTo) : Future.succeededFuture(null);
    return mdFrom
        .compose(x -> mdTo)
        .compose(x -> {
          return options.getDepCheck()
              ? enableAndDisableCheck(tenant, mdFrom.result(), mdTo.result())
              : Future.succeededFuture();
        })
        .compose(x -> enableAndDisableModule(tenant, options, mdFrom.result(), mdTo.result(), pc));
```

okapi-core/src/test/java/org/folio/okapi/MainDeployTest.java
```diff
@@ -22,30 +22,28 @@
 public class MainDeployTest {
 
   private final Logger logger = OkapiLogger.get();
-  private final int port = 9230;
-  private Async async;
-  private Vertx vertx;
-  private RamlDefinition api;
+  private static final int port = 9230;
+  // private Async async;
```
julianladisch, line 26: The comment should be removed.

okapi-core/src/test/java/org/folio/okapi/MainDeployTest.java
```diff
@@ -22,30 +22,28 @@
...
+    System.setProperty("port", Integer.toString(port));
   }
 
   @After
   public void tearDown(TestContext context) {
     System.setProperty("port", ""); // disable port by emptying it
-    if (vertx != null) {
-      vertx.close(context.asyncAssertSuccess());
-      vertx = null;
-    }
```
julianladisch, line 48: `@After` can be changed to `@AfterClass`, setProperty can be changed to clearProperty.

okapi-core/src/test/java/org/folio/okapi/MainDeployTest.java
```diff
@@ -236,87 +189,53 @@ public void testClusterMode(TestContext context) {
...
-
+    d1.init(args, context.asyncAssertSuccess(vertx -> {
       MainDeploy d2 = new MainDeploy();
-      d2.init(args, res2 -> {
-        Assert.assertTrue("d2 " + res2.cause(), res2.failed());
-        async.complete();
-      });
-    });
+      d2.init(args, context.asyncAssertFailure(
+          x -> vertx.close(context.asyncAssertSuccess(y -> async.countDown()))));
```
julianladisch, line 236: countDown() should be replaced by complete()

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: ok

okapi-core/src/test/java/org/folio/okapi/MainDeployTest.java
adamdickmeiss: yep. And @Before can be removed entirely.

okapi-core/src/test/java/org/folio/okapi/MainDeployTest.java
adamdickmeiss: Jesus!. Yes.

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
```diff
@@ -2551,122 +2501,234 @@ public void testSemVer(TestContext context) {
...
+
+  private Future<String> redeploy(TestContext context) {
+    Async async = context.async();
+    Promise<String> promise = Promise.promise();
+    redeploy().onComplete(res -> {
+      promise.handle(res);
+      async.complete();
+    });
+    async.await();
+    return promise.future();
```
julianladisch, line 2532:
```suggestion
  private Future<String> redeploy(TestContext context) {
    Async async = context.async();
    Future<String> future = redeploy().onComplete(res -> async.complete());
    async.await();
    return future;
```

okapi-test-header-module/src/test/java/HeaderModuleTest.java
```diff
@@ -49,52 +49,69 @@ public void tearDown(TestContext context) {
...
+    headers.put(XOkapiHeaders.TENANT, "testlib");
+    JsonObject perm = new JsonObject("{\"k\": \"v\"}");
+    {
+      Async async = context.async();
+      OkapiClient cli = new OkapiClient(URL, vertx, headers);
+      cli.post("/_/tenantPermissions", "{", res1 -> {
+        context.assertTrue(res1.failed());
+        cli.post("/_/tenantPermissions", perm.encode(), res2 -> {
+          context.assertTrue(res2.succeeded());
+          context.assertEquals("GET test-header-module /_/tenantPermissions 200 -",
```
julianladisch, line 88: This is from a POST request, MainVerticle.java should write "POST test-header-module ..." into the header.

okapi-core/src/test/java/org/folio/okapi/MainDeployTest.java
adamdickmeiss: yes

okapi-test-header-module/src/test/java/HeaderModuleTest.java
```diff
@@ -49,52 +49,69 @@ public void tearDown(TestContext context) {
...
+      Async async = context.async();
+      headers.replace(XOkapiHeaders.TENANT, "other");
+      OkapiClient cli = new OkapiClient(URL, vertx, headers);
+      cli.get("/permResult", res -> {
+        context.assertNull(res.result());
+        cli.close();
+        async.complete();
+      });
+      async.await();
+    }
```
julianladisch, line 115: Splitting this big method into smaller ones and using rest-assured could improve readability. It's up to you if you want to change it or merge as is.

okapi-test-header-module/src/test/java/HeaderModuleTest.java
adamdickmeiss: myPermissionHandle hardcodes GET.. I'll fix that.

julianladisch (approved):
Nice work, good code cleanup, thanks!

okapi-test-header-module/src/test/java/HeaderModuleTest.java
adamdickmeiss: ok. I'll do it. Not pretty.

PR #1037, opened 2020-11-06
OKAPI-944 failed tests on mac and ec2

hjiebsco:
Made two small changes to get failed tests pass. For EC2, the connection refused message does not match perfectly. For mac, it turns out the NuProcess returns the sh pid but not the Java pid. So the NuProcess object is not null and is not running, the tryConnect loop does not capture this. Extending the timer wait time to let tryConnect succeed as a work around.

After updating Docker on my Mac, I start to see lots of Ryuk connection error. Updating testcontains version fixed that. There was an issue reported for testcontainer. See https://github.com/testcontainers/testcontainers-java/pull/3159

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
@@ -149,7 +149,7 @@ public class ProcessModuleHandle extends NuAbstractProcessHandler implements Mod
     Promise<Void> promise = Promise.promise();
     // time to wait for process status.. when a port is present (always in real life)..
     // The waitReady will check if process eventually starts listening on port
-    vertx.setTimer(port == 0 ? 3000 : 1000, timerRes -> {
+    vertx.setTimer(3000, timerRes -> {
       if (process.isRunning() || exitCode == 0) {
         promise.complete();
         return;
@@ -198,7 +198,7 @@ public class ProcessModuleHandle extends NuAbstractProcessHandler implements Mod
     Promise<Void> promise = Promise.promise();
     // time to wait for process that shuts down service.. when a port is present (always in prod)
     // The waitPortClose will wait for service to shut down
-    vertx.setTimer(port == 0 ? 3000 : 1000, timerRes -> {
+    vertx.setTimer(3000, timerRes -> {
       if (pp.isRunning() || exitCode == 0) {
         promise.complete();
         return;
--- a/okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
@@ -149,7 +149,7 @@ public class ProcessModuleHandle extends NuAbstractProcessHandler implements Mod
     Promise<Void> promise = Promise.promise();
     // time to wait for process status.. when a port is present (always in real life)..
     // The waitReady will check if process eventually starts listening on port
-    vertx.setTimer(3000, timerRes -> {
+    vertx.setTimer(port == 0 ? 3000 : 1000, timerRes -> {
       if (process.isRunning() || exitCode == 0) {
         promise.complete();
         return;
@@ -198,7 +198,7 @@ public class ProcessModuleHandle extends NuAbstractProcessHandler implements Mod
     Promise<Void> promise = Promise.promise();
     // time to wait for process that shuts down service.. when a port is present (always in prod)
     // The waitPortClose will wait for service to shut down
-    vertx.setTimer(3000, timerRes -> {
+    vertx.setTimer(port == 0 ? 3000 : 1000, timerRes -> {
       if (pp.isRunning() || exitCode == 0) {
         promise.complete();
         return;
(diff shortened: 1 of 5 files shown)


okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
```diff
@@ -149,7 +149,7 @@ private NuProcess launch(EnvEntry[] env, String [] command) {
     Promise<Void> promise = Promise.promise();
     // time to wait for process status.. when a port is present (always in real life)..
     // The waitReady will check if process eventually starts listening on port
-    vertx.setTimer(port == 0 ? 3000 : 1000, timerRes -> {
+    vertx.setTimer(3000, timerRes -> {
```
adamdickmeiss, line 152: How do you conclude that the NuPocess is not returning the right process? As far as I can tell it reaches waitReady as it should, but the Test doesn't wait enough time.. So I think we should just increase waitIterations in ProcessModuleHandleTest, say, from 15 to 20 (desc.setWaitIterations(15)).

The test runs a Bourne Shell script and that's the exit code we check (also on Linux).

Increasing from 1 to 3 seconds will make ALL process deployments 2 seconds slower.

okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
```diff
@@ -198,7 +198,7 @@ private NuProcess launch(EnvEntry[] env, String [] command) {
     Promise<Void> promise = Promise.promise();
     // time to wait for process that shuts down service.. when a port is present (always in prod)
     // The waitPortClose will wait for service to shut down
-    vertx.setTimer(port == 0 ? 3000 : 1000, timerRes -> {
+    vertx.setTimer(3000, timerRes -> {
       if (pp.isRunning() || exitCode == 0) {
```
adamdickmeiss, line 202: Same here.. Increasing waitIterations is a better solution, IMHO.

okapi-core/src/test/java/org/folio/okapi/service/impl/DockerModuleHandleTest.java
```diff
@@ -511,7 +511,7 @@ public void testDockerVersionAtLocal(TestContext context) {
     {
       Async async = context.async();
       dh.pullImage().onComplete(context.asyncAssertFailure(res -> {
-        context.assertTrue(res.getMessage().contains("9231: connect: connection refused"), res.getMessage());
+        assertThat(res.getMessage()).contains("9231").contains("connection refused");
```
adamdickmeiss, line 514: Yes

pom.xml
```diff
@@ -133,7 +133,7 @@
       <dependency>
         <groupId>org.testcontainers</groupId>
         <artifactId>testcontainers-bom</artifactId>
-        <version>1.14.3</version>
+        <version>1.15.0</version>
```
adamdickmeiss, line 136: Nice. Great that was fixed and we don't have to use "snapshot" or similar.

adamdickmeiss (changes requested):
Looks good. Thx

okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
hjiebsco: I tried that initially and it did not work. I added some debug statements and found out the PID difference. For example:
```
Starting test: testCmdlineOk
07:59:13 [] [] [] [] INFO  Messages             Loading messages from /infra-messages/Messages_en.properties
07:59:15 [] [] [] [] INFO  TcpPortWaiting       tryConnect() host localhost port 9231 count 0
res suceeded: false
process is null? false
process id 52222
process is running? false
```
The actual Java process
```
hji              52223   0.0  0.6 10281380  98932   ??  S     7:59AM   0:02.48 /usr/bin/java -DpidFile=test-module.pid -Dport=9231 -jar ../okapi-test-module/target/okapi-test-module-fat.jar
```

adamdickmeiss: The PID is a red herring..  Not related.. The PID is the Bourne Shell on my system too. The error message is coming from the code that waits for the port to become ready.. Please increase it to 20.

hjiebsco: It does not work. See below console output for failed testCmdlineOk test. The reason it does not work is here https://github.com/folio-org/okapi/blob/master/okapi-core/src/main/java/org/folio/okapi/util/TcpPortWaiting.java#L41-L53. 
When res is false, proces is not null and not running (because of PID mismatch), the try maxIteration loop does not kick in. BTW, in AWS ec2 linux evn, the PID from sh and Java is the same.  But on my macOS that is not the case.
```
Starting test: testCmdlineOk
09:39:23 [] [] [] [] INFO  Messages             Loading messages from /infra-messages/Messages_en.properties
09:39:25 [] [] [] [] INFO  TcpPortWaiting       tryConnect() host localhost port 9231 count 0
maxIterations is 20
res suceeded: false
process is null? false
process id 54052
process is running? false
```

adamdickmeiss: You are right .. isRunning is the problem .. So we should not even bother to check it when the process spawned is a Bourne shell .. (cmdLineStart, cmdLineStop).

hjiebsco: Made changes suggested by @adamdickmeiss

adamdickmeiss (approved):

PR #1040, opened 2020-11-10
Empty dockerRegistries same as no dockerRegistries OKAPI-945

adamdickmeiss:


julianladisch:
I disagree with the change. For details see
* https://issues.folio.org/browse/OKAPI-945
* Pull request #1042: `OKAPI-945: Document dockerRegistries [] and [ {} ]`

adamdickmeiss:
I'm happy with either way.. Just note that current behavior obviously has caused confusion.

PR #1041, opened 2020-11-10
Support both "http.port" and "port" property variables

julianladisch:
"http.port" is the command-line option name that all RMB based
modules support:
https://github.com/folio-org/raml-module-builder#command-line-options

For consistency Okapi should also support this.

The guide.md should show "http.port" because most modules
are RMB based and support only this form.

### DIFF
--- a/dist/okapi.sh
+++ b/dist/okapi.sh
@@ -121,9 +121,19 @@ parse_okapi_conf()  {
       OKAPI_JAVA_OPTS+=" -Dhost=${host}"
    fi
 
-   # configure okapi port
-   if [ "$port" ]; then
-      OKAPI_JAVA_OPTS+=" -Dport=${port}"
+   # configure okapi http port
+   # Environment variables with periods/dots in their names are deprecated
+   # because a period is not POSIX compliant and therefore some shells, notably,
+   # the BusyBox /bin/sh included in Alpine Linux, strip them:
+   # https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html
+   # https://github.com/docker-library/docs/tree/master/openjdk#environment-variables-with-periods-in-their-names
+   # https://wiki.ubuntu.com/DashAsBinSh
+   # https://wiki.ubuntu.com/DashAsBinSh/Spec
+   # Java property variable names allow periods/dots.
+   if [ "$http_port" ]; then
+      OKAPI_JAVA_OPTS+=" -Dhttp.port=${http_port}"
+   elif [ "$port" ]; then
+      OKAPI_JAVA_OPTS+=" -Dhttp.port=${port}"
    fi
 
    # configure module port range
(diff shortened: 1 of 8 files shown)


okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
```diff
@@ -83,7 +83,8 @@ public void init(Vertx vertx, Context context) {
     super.init(vertx, context);
 
     JsonObject config = context.config();
-    port = Integer.parseInt(Config.getSysConf("port", "9130", config));
+    String portString = System.getProperty("http.port", "9130");
+    port = Integer.parseInt(Config.getSysConf("port", portString, config));
     String okapiVersion2 = Config.getSysConf("okapiVersion", null, config);
```
adamdickmeiss, line 88: So port is only supported as property. Not as Json configuration. Is that intended?

julianladisch: No, good catch!
Fixed.

dist/okapi.sh
```diff
@@ -121,9 +121,19 @@ parse_okapi_conf()  {
...
+   # https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html
+   # https://github.com/docker-library/docs/tree/master/openjdk#environment-variables-with-periods-in-their-names
+   # https://wiki.ubuntu.com/DashAsBinSh
+   # https://wiki.ubuntu.com/DashAsBinSh/Spec
+   # Java property variable names allow periods/dots.
+   if [ "$http_port" ]; then
+      OKAPI_JAVA_OPTS+=" -Dhttp.port=${http_port}"
+   elif [ "$port" ]; then
+      OKAPI_JAVA_OPTS+=" -Dhttp.port=${port}"
    fi
```
adamdickmeiss, line 137: I'm not sure I follow. This is a value, not a name. Hopefully no system tampers with values. So I don't think we should excuse anything here.

adamdickmeiss: You mean http_port of course. ok.

adamdickmeiss (approved):

adamdickmeiss:
In retrospect, I wished we'd go for http_port everywhere. For Okapi other options are with underscore.. Eg postgres_port . RMB also started with periods in DB variables.. That was then changed to be with _..

hjiebsco (commented):
Personally I would strongly prefer Okapi and all modules can be consolidated to use a single property variable `port` to be consistent, concise, and cover modules that do not use http, for example edge-sip2.

julianladisch:
Blockes by https://issues.folio.org/browse/FOLIO-2876 "Consistently use -Dport or -Dhttp.port for Okapi HTTP port"

hjiebsco (approved):
Left a comment but I am fine with it if team thinks `http.port` is better than `port`.

PR #1042, opened 2020-11-11
OKAPI-945: Document dockerRegistries [] and [ {} ]

julianladisch:
[] is zero registries and disables pull requests.
[{}] is unauthenticated DockerHub.

### DIFF
--- a/dist/okapi.sh
+++ b/dist/okapi.sh
@@ -37,6 +37,13 @@ parse_okapi_conf()  {
          OKAPI_JAVA_OPTS+=" -Dstorage=inmemory"
       fi
 
+      # Validate JSON of $docker_registries, undefined is valid
+      echo "$docker_registries" | jq empty
+      if [ $? -ne 0 ]; then
+         echo "Exiting because \$docker_registries is invalid"
+         exit 3
+      fi
+
       # create runtime configuration file for sensitive information
       # Store auth here rather than exposing on command line
       if [ "$storage" == "postgres" ] || [ "$docker_registries" ]; then
@@ -46,7 +53,7 @@ parse_okapi_conf()  {
          chmod 600 "${PID_DIR}/okapi-runtime.conf"
          # include postgres connection arguments only if defined
          jq -n --arg WARNING "AUTOMATICALLY CREATED FILE, DO NOT EDIT!"    \
-               --argjson dockerRegistries  ${docker_registries:-[\{\}]}    \
+               --argjson dockerRegistries  "${docker_registries:-[\{\}]}"  \
                --arg postgres_host         "${postgres_host}"              \
                --arg postgres_port         "${postgres_port}"              \
                --arg postgres_username     "${postgres_username}"          \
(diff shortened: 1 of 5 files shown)


julianladisch:
If this pull request is accepted it obsoletes the other pull request:
* Empty dockerRegistries same as no dockerRegistries OKAPI-945 #1040

adamdickmeiss (approved):
I'm fine with this.

okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
```diff
@@ -37,6 +37,10 @@
 @java.lang.SuppressWarnings({"squid:S1192"})
 public class DockerModuleHandle implements ModuleHandle {
 
+  static final String DOCKER_REGISTRIES_EMPTY_LIST =
+      "dockerRegistries=[] contains no registry and disables docker pull. "
+          + "For unauthenticated pull from DockerHub use [ {} ].";
+
   private final Logger logger;
```
adamdickmeiss, line 44: I think it's better to assign a message for this by adding an entry in okapi-core/src/main/resources/infra-messages/Messages_en.properties

julianladisch: https://github.com/folio-org/okapi/tree/master/okapi-core/src/main/resources/infra-messages hasn't been used since 2018. And there isn't any translation yet. I would rather propose to remove all infra-messages to improve code readability.

adamdickmeiss: Are you kidding?

adamdickmeiss: FWIW, let's must merge this now and have this discussion later.. The messages returned IMHO should be accommodated with the message ID .. Eg (OKAPI-12345) messsage .. It serves as a nice unique way of telling what has returned the message. Of course, the log might have backtraces and all.. But for a system user, it's nice to be able to spot WHAT has actually generated some message .. Okapi? RMB ? Or code from a module?

wafschneider (approved):
Reviewing the documentation only, it is much improved and clarifies the behavior very well.

wafschneider:
Added a commit to align the default behavior in okapi.sh with the documented behavior. The `--argjson` option to jq requires JSON input, so set the default to be `[{}]` instead of `[]`. Makes no operational difference, but might be less confusing if anyone ever looks at the script and wonders how the default gets used.

dcrossleyau (approved):
I too only investigate documentation. Good. Thanks.

dist/okapi.sh
```diff
@@ -46,7 +53,7 @@ parse_okapi_conf()  {
          chmod 600 "${PID_DIR}/okapi-runtime.conf"
          # include postgres connection arguments only if defined
          jq -n --arg WARNING "AUTOMATICALLY CREATED FILE, DO NOT EDIT!"    \
-               --argjson dockerRegistries  ${docker_registries:-[\{\}]}    \
+               --argjson dockerRegistries  "${docker_registries:-[\{\}]}"  \
```
wafschneider, line 56: @julianladisch in my testing using Ubuntu Focal, if `$docker_registries` is undefined, this fails if quoted (though it works if `$docker_registries` is defined). I confess I'm not sure why. Did it work in your testing?

wafschneider:
@julianladisch excellent ideas for improving okapi.sh and okapi.conf, see my comment above about a possible bug.

dist/okapi.sh
julianladisch: Good catch!
This is correct: `"${docker_registries:-[{\}]}"`
When using quotes only the closing } needs to be masked.

```
$ docker run --rm=true -it ubuntu:focal
root@e4921c64df45:/# apt-get update; apt-get install -y jq
root@e4921c64df45:/# jq -n --argjson dockerRegistries "${docker_registries:-[{\}]}" "{ \$dockerRegistries }"                                   {
  "dockerRegistries": [
    {}
  ]
}
```

PR #1043, opened 2020-11-11
Honor explicit dependencies (notion: fixed) OKAPI-946

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/bean/TenantModuleDescriptor.java
```diff
@@ -33,6 +33,10 @@
     pending, deploy, invoke, undeploy, done
   }
 
+  // if true, it specifies that the id is specified by user and can not be changed
```
julianladisch, line 36: This comment should be reworded, converted into javadoc and attached to the public methods getFixed() and setFixed(Boolean fixed).

`Returns true if the version is specified by user and the product can not be upgraded or downgraded.`

okapi-core/src/main/java/org/folio/okapi/bean/TenantModuleDescriptor.java
```diff
@@ -33,6 +33,10 @@
...
+  @JsonIgnore
+  private Boolean fixed;
```
julianladisch, line 38: I don't see the need why this should have three values: true, false, null. Can a boolean (with true and false only) been used instead?

julianladisch (approved):
Looks good, thanks!

okapi-core/src/main/java/org/folio/okapi/bean/TenantModuleDescriptor.java
adamdickmeiss: True. I'll change that.

PR #1046, opened 2020-11-16
Forward Content-Encoding header OKAPI-936

adamdickmeiss:
Fix for OKAPI-936 Request logging can send compressed data to the filter. Until now only the Content-Type was forwarded from request-response pipeline.. request-only, headers should NOT forward this as the response is thrown away.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -857,11 +857,8 @@ public class ProxyService {
           makeTraceHeader(mi, res.statusCode(), pc);
           relayToRequest(res, pc, mi);
           for (String header : FORWARD_HEADERS) {
-            String value = res.getHeader(header);
-            request.headers().remove(header);
-            if (value != null) {
-              request.headers().set(header, value);
-            }
+            MultiMap headers = request.headers();
+            request.headers().set(header, res.getHeader(header));
           }
           storeResponseInfo(pc, mi, res);
           res.pause();
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -85,6 +85,9 @@ public class ProxyService {
       Comparator.comparing((ModuleInstance a) -> a.getRoutingEntry().getPhaseLevel());
   private final TokenCache tokenCache;
 
+  // request + response HTTP headers that are forwarded in the pipeline
+  private static final String [] FORWARD_HEADERS = new String [] { "Content-Type", "Content-Encoding"};
+
   /**
    * Construct Proxy service.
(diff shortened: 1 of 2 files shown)


okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -844,21 +850,24 @@ private void proxyRequestResponse(Iterator<ModuleInstance> it,
...
           relayToRequest(res, pc, mi);
-          final String ct = res.getHeader("Content-Type");
-          if (ct != null) {
-            ctx.request().headers().set("Content-Type", ct);
+          for (String header : FORWARD_HEADERS) {
+            String value = res.getHeader(header);
+            request.headers().remove(header);
+            if (value != null) {
+              request.headers().set(header, value);
+            }
```
hjiebsco, line 864: Seems you can combine those to one line `request.headers().set(header, res.getHeader(header));`

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4344,4 +4354,170 @@ public void testManyModules(TestContext context) throws IOException {
...
+    Assert.assertTrue(
+        "raml: " + c.getLastReport().toString(),
+        c.getLastReport().isEmpty());
+
+    given().header("X-Okapi-Tenant", okapiTenant)
+        .header("Content-Type", "text/plain")
+        .body("Okapi").post("/echo")
+        .then().statusCode(200)
+        .header("Content-Type", "text/plain; charset=ISO-8859-1")
+        .body(equalTo("Okapi"));
```
hjiebsco, line 4476: Consider to add a line to check non-existing gzip header `.header("Content-Encoding", nullValue())`

hjiebsco (approved):
Left two minor comments.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: Can I? Setting null is the same as not setting it at all?

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
adamdickmeiss: ok

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
hjiebsco: The doc says `If there is an existing header with the same name, it is removed. Setting a null value removes the entry.` So I think it is yes. Please verify though.

adamdickmeiss: Right you are.. Thx

hjiebsco (approved):

PR #1053, opened 2020-12-05
OKAPI-875 async tenant init

adamdickmeiss:
PR for version 2.0 of _tenant interface. It will use the paths for POST, GET, DELETE as provided. Version 1 series are still supported and there should be no changes in behavior. So this is a pure extension as far as Okapi is concerned.

Should a Location be returned at runtime and yet the module descriptor only says 1.1 / 1.2 , Okapi issues an error. That is maintainer forgets https://github.com/folio-org/raml-module-builder/pull/815 .

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -48,7 +48,6 @@ managing and running microservices.
     * [Instrumentation](#instrumentation)
 * [Module Reference](#module-reference)
     * [Life cycle of a module](#life-cycle-of-a-module)
-    * [Tenant Interface](#tenant-interface)
     * [HTTP](#http)
 
 ## Introduction
@@ -3204,7 +3203,7 @@ In FOLIO two such parameters are widely recognized:
  * `loadReference` with value `true` loads reference data.
  * `loadSample` with value `true` loads sample data.
 
-### Tenant Interface
+#### Tenant Interface definitions
 
 A module supporting 1.1/1.2 of the `_tenant` interface should use this
 snippet in the module descriptor:
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -3206,20 +3206,21 @@ In FOLIO two such parameters are widely recognized:
 
 ### Tenant Interface
 
-A module supporting 1.1/1.2 of the `_tenant` interface should use this snippet
-in the module descriptor:
+A module supporting 1.1/1.2 of the `_tenant` interface should use this
+snippet in the module descriptor:
 
 ```
    "id" : "_tenant",
    "version" : "1.2",
    "interfaceType" : "system",
    "handlers" : [ {
-     "methods" : [ "POST", "DELETE" ],
-     "pathPattern" : "/_/tenant"
-    }, {
-     "methods" : [ "POST" ],
-     "pathPattern" : "/_/tenant/disable"
-    } ]
+       "methods" : [ "POST", "DELETE" ],
+       "pathPattern" : "/_/tenant"
+      }, {
+       "methods" : [ "POST" ],
+       "pathPattern" : "/_/tenant/disable"
+      }
+   ]
 ```
 The corresponding RAML definition:
 [tenant.raml](https://github.com/folio-org/raml/blob/tenant_interface_1_2/ramls/tenant.raml)
@@ -3231,12 +3232,13 @@ Snippet of version 2.0 of the `_tenant` interface:
    "version" : "2.0",
    "interfaceType" : "system",
    "handlers" : [ {
-     "methods" : [ "POST" ],
-     "pathPattern" : "/_/tenant"
-    }, {
-     "methods" : [ "GET", "DELETE" ],
-     "pathPattern" : "/_/tenant/{id}"
-    } ]
+       "methods" : [ "POST" ],
+       "pathPattern" : "/_/tenant"
+      }, {
+       "methods" : [ "GET", "DELETE" ],
+       "pathPattern" : "/_/tenant/{id}"
+      }
+   ]
 ```
 The corresponding RAML definition:
 [tenant.raml](https://github.com/folio-org/raml/blob/raml1.0/ramls/tenant.raml)
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -3001,7 +3001,7 @@ the module, and are no longer used.
 
 For the [specifics](#web-service), see under
 `.../okapi/okapi-core/src/main/raml/raml-util` the files
-`ramls/tenant.raml` and `schemas/moduleInfo.schema`.  The
+`ramls/tenant.raml` and `schemas/tenantAttributes.schema`.  The
 okapi-test-header-module has a very trivial implementation of this,
 and the moduleTest shows a module Descriptor that defines this
 interface.
@@ -3142,9 +3142,19 @@ during the operation.
 
 Add 'loadSample=true` to parameters, to load sample data as well.
 
+If the module supports version 2 of the `_tenant` interface it should return status 201 with
+a `Location` header. This signals that the tenant job has started. Use a GET request on the
+Location path returned to poll for the completion of the tenant job with the
+
+    curl -HX-Okapi-url:http://localhost:8081 -HX-Okapi-Tenant:testlib \
+      -HContent-Type:application/json "-HAccept:*/*" \
+      http://localhost:8081/path
+
+(substitute above path with the Location returned)
+
 #### Upgrading
 
-When a module gets upgraded to a new version, it happens separately
+When a module is upgraded to a new version, it happens separately
 for each tenant.  Some tenants may not wish to upgrade in the middle
 of a busy season, others may want to have everything in the latest
 version. The process starts by Okapi deploying the new version of the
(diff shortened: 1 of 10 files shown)


marcjohnson-kint:
> Should a Location be returned at runtime and yet the module descriptor only says 1.1 / 1.2 , Okapi will warn and continue to use it as if it is version 2.

Does that mean that Okapi will effectively ignore the interface version stated by the module and assume that it was wrong?

adamdickmeiss:
Yes. Older RMB will not return a Location so it's a pretty safe bet. And I know how easy it is to forget to update the MD.

marcjohnson-kint:
> Yes. Older RMB will not return a Location so it's a pretty safe bet. And I know how easy it is to forget to update the MD.

I would've thought we would want Okapi to fail in these circumstances in order to alert the module owners that the Tenant API was either incorrectly implemented or declared.

Otherwise, how would these mistakes get corrected if Okapi compensates for them automatically?

adamdickmeiss:
ok. Making it a failure instead.

okapi-core/src/main/java/org/folio/okapi/bean/ModuleInstance.java
```diff
@@ -44,6 +44,10 @@ public ModuleInstance(ModuleDescriptor md, RoutingEntry re,
     this.withRetry = false;
   }
 
+  public void substPath(String pattern, String value) {
+    path = path.replace(pattern, value);
+  }
+
```
julianladisch, line 50: Are there plans for other substitutions or can we use
```
/**
 * Substitute {id} in path
 */
public void substPathId(String id) {
  path = path.replace("{id}", id);
}
```

adamdickmeiss: Not now. Can change it, if you like.

doc/guide.md
```diff
@@ -3188,29 +3196,51 @@ etc.  also load sets of reference data. This can be controlled by
...
+   "version" : "2.0",
+   "interfaceType" : "system",
+   "handlers" : [ {
+       "methods" : [ "POST" ],
+       "pathPattern" : "/_/tenant"
+      }, {
+       "methods" : [ "GET", "DELETE" ],
+       "pathPattern" : "/_/tenant/{id}"
+      }
+   ]
```
julianladisch, line 3240: Show we add `"permissionsRequired" : [ ]`?

adamdickmeiss: We could, but kind of redundant. They aren't required for system interfaces.

doc/guide.md
```diff
@@ -3142,9 +3138,19 @@ during the operation.
 
 Add 'loadSample=true` to parameters, to load sample data as well.
 
+If the module supports version 2 of the `_tenant` interface it should return status 201 with
+a `Location` header. This signals that the tenant job has started. Use a GET request on the
+Location path returned to poll for the completion of the tenant job with the
+
+    curl -HX-Okapi-url:http://localhost:8081 -HX-Okapi-Tenant:testlib \
+      -HContent-Type:application/json "-HAccept:*/*" \
```
julianladisch, line 3146:
```suggestion
    curl -H "X-Okapi-url: http://localhost:8081" -H "X-Okapi-Tenant: testlib" \
      -H "Content-Type: application/json" -H "Accept: */*" \
```

doc/guide.md
```diff
@@ -3142,9 +3138,19 @@ during the operation.
 
 Add 'loadSample=true` to parameters, to load sample data as well.
 
+If the module supports version 2 of the `_tenant` interface it should return status 201 with
+a `Location` header. This signals that the tenant job has started. Use a GET request on the
```
julianladisch, line 3142:
```suggestion
a

    Location: /_/tenant/6da99bac-457b-499f-89a4-34f4da8e9be8
    
header. This signals that the tenant job has started. Use a GET request on the
```

doc/guide.md
```diff
@@ -3142,9 +3138,19 @@ during the operation.
...
+Location path returned to poll for the completion of the tenant job with the
+
+    curl -HX-Okapi-url:http://localhost:8081 -HX-Okapi-Tenant:testlib \
+      -HContent-Type:application/json "-HAccept:*/*" \
+      http://localhost:8081/path
```
julianladisch, line 3147:
```suggestion
      http://localhost:8081/_/tenant/6da99bac-457b-499f-89a4-34f4da8e9be8
```

okapi-core/src/main/java/org/folio/okapi/bean/ModuleInstance.java
julianladisch: I like it. Otherwise the task is split: Call to replace in ModuleInstance, substitution string "{id}" in TenantManager.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -1193,6 +1193,40 @@ private static DeploymentDescriptor pickInstance(List<DeploymentDescriptor> inst
...
 
+  private Future<OkapiClient> doCallSystemInterface2(
+      MultiMap headersIn, String tenantId, String authToken,
+      ModuleInstance inst, String modPerms, String request) {
+
+    Map<String, String> headers = sysReqHeaders(headersIn, tenantId, authToken, inst, modPerms);
+    headers.put(XOkapiHeaders.URL_TO, inst.getUrl());
+    logger.debug("syscall begin {} {}{}", inst.getMethod(), inst.getUrl(), inst.getPath());
+    OkapiClient cli = new OkapiClient(this.httpClient, inst.getUrl(), vertx, headers);
+    String reqId = inst.getPath().replaceFirst("^[/_]*([^/]+).*", "$1");
```
julianladisch, line 1204:
```suggestion
    // first path segment, but skip leading /_
    String reqId = inst.getPath().replaceFirst("^(/_)?/([^/]+).*", "$2");
```

okapi-core/src/test/java/org/folio/okapi/InstallTest.java
```diff
@@ -1119,7 +1047,325 @@ public void installTenantInit(TestContext context) {
...
+            + "  \"id\" : \""+ module + "\"," + LS
+            + "  \"action\" : \"enable\"" + LS
+            + "} ]"))
+        .extract().response();
+    Assert.assertTrue(
+        "raml: " + c.getLastReport().toString(),
+        c.getLastReport().isEmpty());
+    String locationInstallJob = r.getHeader("Location");
+
+    String path = locationInstallJob.substring(locationInstallJob.indexOf("/_/"));
```
julianladisch, line 1112: `indexOf("/_/")` is always 0.
```suggestion
    String path = r.getHeader("Location");
```

doc/guide.md
julianladisch: Agreed.

doc/guide.md
adamdickmeiss: More typing, but I guess it's more readable.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: This replaceFirst was not part of PR (only due to refactorings)..

The initial replaceFirst would make `tenant` from path  `/_/tenant`. Your suggestion ends up with an empty string.

okapi-core/src/test/java/org/folio/okapi/InstallTest.java
adamdickmeiss: LOL. Yes.. I was probably thinking that Location could be absolute URL!

doc/guide.md
marcjohnson-kint: I thought it was agreed that all endpoints in module descriptors *should* declare an empty `modulePermissions` property to make it explicit that a conscious decision had been made.

@craigmcnally @skoczko Did I misunderstand that guidance?

okapi-core/src/main/java/org/folio/okapi/bean/ModuleInstance.java
adamdickmeiss: Done.

okapi-core/src/test/java/org/folio/okapi/InstallTest.java
marcjohnson-kint: > LOL. Yes.. I was probably thinking that Location could be absolute URL!

Is this the location for the status of the install task in Okapi or the location of the status of an individual module upgrade?

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
julianladisch: `System.out.println("/_/tenant".replaceFirst("^(/_)?/([^/]+).*", "$2"));`
outputs
`tenant`

For all unit tests my suggestion returns the same result as before.

julianladisch (approved):
Looks good, thanks!

MikhailFokanov (approved):
Looks good.

PR #1056, opened 2020-12-10
Update Okapi systemd service configuration

wafschneider:
OKAPI-966. This change is required because of a change to the docker-ce Debian package.


julianladisch (approved):
Looks good, thanks!

adamdickmeiss (approved):

PR #1057, opened 2020-12-11
OKAPI-947 - Updates for static permissions migration

craigmcnally:
## Purpose
Implement changes to facilitate static permission migration.  For details, see https://wiki.folio.org/display/DD/Migration+of+Static+Permissions+Upon+Upgrade?src=contextnavpagetreemode

Also see:  https://issues.folio.org/browse/OKAPI-947

## Approach
* Add a `renamedFrom` field to the module descriptor permission schema indicating a permission's previously used name(s)
* Adjust the payload for the _tenantPermissions call to include:
  * `moduleIdFrom` - the name/version of the module currently in use/being upgraded from (if applicable)
  * `permsFrom` - the list of permissions defined by the version of the module currently in use (if applicable)
  * `permsTo` - renamed from `perms` for clarity. 
  * `moduleIdTo` - renamed from `moduleId` for clarity.
* When upgrading mod-permissions, call the _tenantPermissions API for each module enabled for the tenant, resulting in the permissions being refreshed.  This is already done when first enabling mod-permissions install, but is now also done when upgrading it.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
@@ -659,13 +659,16 @@ public class TenantManager implements Liveness {
       pl = new PermissionListV1(moduleTo, mdTo.getPermissionSets());
     } else if (permIntVer.equals("1.1")) {
       pl = new PermissionListV1(moduleTo, mdTo.getExpandedPermissionSets());
-    } else {
+    } else if (permIntVer.equals("2.0")) {
       if (mdFrom != null) {
         pl = new PermissionListV2(moduleTo, mdFrom.getId(), mdTo.getExpandedPermissionSets(),
             mdFrom.getExpandedPermissionSets());
       } else {
         pl = new PermissionListV2(moduleTo, mdTo.getExpandedPermissionSets());
       }
+    } else {
+      return Future.failedFuture(new OkapiError(ErrorType.ANY,
+          "Unknown version of _tenantPermissions interface in use " + permIntVer + "."));
     }
     String pljson = Json.encodePrettily(pl);
     logger.debug("tenantPerms Req: {}", pljson);
--- a/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
@@ -667,7 +667,7 @@ public class TenantManager implements Liveness {
         pl = new PermissionListV2(moduleTo, mdTo.getExpandedPermissionSets());
       }
     } else {
-      return Future.failedFuture(new OkapiError(ErrorType.ANY,
+      return Future.failedFuture(new OkapiError(ErrorType.USER,
           "Unknown version of _tenantPermissions interface in use " + permIntVer + "."));
     }
     String pljson = Json.encodePrettily(pl);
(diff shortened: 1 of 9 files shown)


adamdickmeiss (changes requested):
There's no check for the version of _tenantPermisssions. So this will not work with existing mod-permissions. Existing supported interface versions: 1.0 and 1.1. New interface version: 2.0.

craigmcnally:
> There's no check for the version of _tenantPermisssions. So this will not work with existing mod-permissions. Existing supported interface versions: 1.0 and 1.1. New interface version: 2.0.
> 
> I'd hate this be a breaking change. Eg Okapi 5 (of course the tenantPermissions is a breaking change). I really think we should aim at supporting both 1 + 2 at the same time.. Any version we release in version 4 series of Okapi should be able to run with modules that also were were running with Okapi 4.0.0.

You're right.  I'll fix this.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -651,16 +648,24 @@ private void fireTimer(Tenant tenant, ModuleDescriptor md, RoutingEntry re, Stri
...
+    } else if (permIntVer.equals("1.1")) {
+      pl = new PermissionListV1(moduleTo, mdTo.getExpandedPermissionSets());
     } else {
-      pl = new PermissionList(moduleTo, mdTo.getExpandedPermissionSets());
+      if (mdFrom != null) {
+        pl = new PermissionListV2(moduleTo, mdFrom.getId(), mdTo.getExpandedPermissionSets(),
+            mdFrom.getExpandedPermissionSets());
+      } else {
+        pl = new PermissionListV2(moduleTo, mdTo.getExpandedPermissionSets());
+      }
```
adamdickmeiss, line 668: Please add check for "2.0". And report failure if no match (return failed future).

craigmcnally: And in the case where we get something other than 1.0, 1.1, or 2.0, also return a failure?  I guess the assumption there is that when/if 2.1 comes along we'll need to make changes here anyway?

craigmcnally: NM I need to read more closely.

adamdickmeiss: Yes. Okapi should report a failure for an unknown interface version.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -659,13 +659,16 @@ private void fireTimer(Tenant tenant, ModuleDescriptor md, RoutingEntry re, Stri
...
       if (mdFrom != null) {
         pl = new PermissionListV2(moduleTo, mdFrom.getId(), mdTo.getExpandedPermissionSets(),
             mdFrom.getExpandedPermissionSets());
       } else {
         pl = new PermissionListV2(moduleTo, mdTo.getExpandedPermissionSets());
       }
+    } else {
+      return Future.failedFuture(new OkapiError(ErrorType.ANY,
+          "Unknown version of _tenantPermissions interface in use " + permIntVer + "."));
     }
```
adamdickmeiss, line 672: I prefer you use ErrorType.USER here so that it returns 400, not 500. I think we should only return 500 when Okapi makes unexpected failures that are "not meant to happen".

adamdickmeiss (approved):
Good. Only that I'd prefer 400 rather than 500.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
craigmcnally: fair enough

adamdickmeiss (approved):

hjiebsco:
I see both `PermissionListV1.java` and `PermissionListV1.java`. Just curious if we considered the possibility of not renaming the existing fields from `moduleId` to `moduleToId` and `perms` to `permsTo`? That could potentially avoid two versions of the file and still be full backwards compatible.

PR #1058, opened 2020-12-14
Publish reload-modules also when disabling module OKAPI-967

adamdickmeiss:
Fix this brain-damage - due to previous behavior when we were caching modules as a whole (for all tenants).. In that case we could ignore when a module was disabled.. In the current design.. of course all changes to modules enabled for a tenant needs to be published to all okapi instances.


julianladisch (approved):

ArgonV (approved):
Built and tested OKAPI-967-missing-reload-when-module-is-disabled branch with deployment of 3 Okapi pods forming a cluster.

Ran my bootstrap-superuser Job - it disabled, and enabled the appropriate modules correctly while also creating the superuser and permissions for it.

PR #1063, opened 2020-12-18
OKAPI-968: Unhandled exception in DockerModuleHandle.request

julianladisch:
Vert.x HttpClient.request doesn't return all exceptions as a failed future but also directly throws some of them.

Example:
```
2020-12-18T22:03:31,099 INFO  DockerModuleHandle   delete container 8ee1e1ec95ecc311d7d603a4d5871ac17dd67c860a48f613ce0c7199e4e4c59b image folioci/mod-gobi:1.
8.0-SNAPSHOT.98
2020-12-18T22:03:31,102 ERROR ?                    Unhandled exception
java.lang.IllegalStateException: Client is closed
        at io.vertx.core.http.impl.HttpClientImpl.checkClosed(HttpClientImpl.java:634) ~[okapi-core-fat.jar:?]
        at io.vertx.core.http.impl.HttpClientImpl.request(HttpClientImpl.java:542) ~[okapi-core-fat.jar:?]
        at io.vertx.core.http.impl.HttpClientImpl.request(HttpClientImpl.java:396) ~[okapi-core-fat.jar:?]
        at io.vertx.core.http.impl.HttpClientImpl.request(HttpClientImpl.java:391) ~[okapi-core-fat.jar:?]
        at org.folio.okapi.service.impl.DockerModuleHandle.request(DockerModuleHandle.java:145) ~[okapi-core-fat.jar:?]
        at org.folio.okapi.service.impl.DockerModuleHandle.deleteUrl(DockerModuleHandle.java:159) ~[okapi-core-fat.jar:?]
        at org.folio.okapi.service.impl.DockerModuleHandle.deleteContainer(DockerModuleHandle.java:175) ~[okapi-core-fat.jar:?]
        at org.folio.okapi.service.impl.DockerModuleHandle.lambda$prepareContainer$18(DockerModuleHandle.java:430) ~[okapi-core-fat.jar:?]
```

Task:
Catch them and pass them as a failed future.


adamdickmeiss (approved):
OK. However, since this method is used in ProxyService I think it would be convenient to just make a wrapper around it as a utility once and for all (that never throws exceptions).

There are many variants of request in Vert.x, but `request(RequestOptions options)` is all we need.

okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
```diff
@@ -136,18 +136,22 @@ static String setupDockerAddress(StringBuilder socketAddress, String u) {
...
-    RequestOptions requestOptions = new RequestOptions()
-        .setMethod(method)
-        .setAbsoluteURI(dockerUrl + url);
-    if (socketAddress != null) {
-      requestOptions.setServer(socketAddress);
-    }
-    return client.request(requestOptions).compose(request -> {
-      if (headers != null) {
-        request.headers().setAll(headers);
+    try {
```
hjiebsco, line 139: Should make the try block smaller to start from line 146?

hjiebsco (commented):
What caused the `java.lang.IllegalStateException: Client is closed` error in PR description? Should we check `client` is still valid before making the http call?

adamdickmeiss:
I'm seeing illegalstateexceotion today on Windows. I never saw it on Ubuntu before. I'll investigate.

okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
julianladisch: Done.

julianladisch:
We don't need to check that HttpClient is still valid. HttpClient already makes this check and throws an IllegalStateException, no need to duplicate this code.

hjiebsco:
> We don't need to check that HttpClient is still valid. HttpClient already makes this check and throws an IllegalStateException, no need to duplicate this code.

Let's see if Adam can find out why caused that exception. `IllegalStateException` is a runtime exception. I was wondering if that particular `HttpClient` instance was accidentally closed somewhere due to incorrect usage or some system bug.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -1201,7 +1201,7 @@ private static DeploymentDescriptor pickInstance(List<DeploymentDescriptor> inst
     Map<String, String> headers = sysReqHeaders(headersIn, tenantId, authToken, inst, modPerms);
     headers.put(XOkapiHeaders.URL_TO, inst.getUrl());
     logger.debug("syscall begin {} {}{}", inst.getMethod(), inst.getUrl(), inst.getPath());
-    OkapiClient cli = new OkapiClient(this.httpClient, inst.getUrl(), vertx, headers);
+    OkapiClient cli = new OkapiClient(httpClient.getHttpClient(), inst.getUrl(), vertx, headers);
```
hjiebsco, line 1204: OkapiClient does not need this new `FuturisedHttpClient`?

julianladisch: OkapiClient wraps HttpClient into WebClient, and WebClient wraps the call into a try-catch-block: https://github.com/vert-x3/vertx-web/blob/4.0.0/vertx-web-client/src/main/java/io/vertx/ext/web/client/impl/HttpContext.java#L305

hjiebsco (approved):

PR #1065, opened 2021-01-08
OKAPI-971: Fail unit tests if docker is not available

julianladisch:
FOLIO switches from embedded-postgresql to testcontainers.org that use
Docker PostgreSQL containers for tests in RMB based modules, RMB and Okapi:
FOLIO-1845

It no longer makes sense to skip tests if Docker is not available.

Tests should fail by default if we cannot run all tests so that any Docker
malfunction on the developer machine and on CI is noticed immediately. It
is misleading if such tests are skipped by default, see comments on OKAPI-970.

DockerModuleHandleTest, ModuleTest and PostgresHandleTest require Docker.

This pull request removes

    @Testcontainers(disabledWithoutDocker = true)

PostgresHandleTest. (The other two files already fail on missing Docker).
And it adjust README.

If there really is a need to run only those tests that don't require
Docker please create a new Jira for this feature. This might be
implemented by adding a `withoutDocker` maven profile to the pom.xml
that can be used in this way:

    mvn clean install -PwithoutDocker

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,11 @@ The Okapi software has the following compile-time dependencies:
 
 * Apache Maven 3.3.x or higher
 
-In addition, the test suite must be able to bind to ports 9230-9239 to
-succeed.
+The test suite has these additional dependencies:
+
+* Docker, for details see https://www.testcontainers.org/supported_docker_environment/
+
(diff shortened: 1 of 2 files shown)


adamdickmeiss (approved):
ok

PR #1066, opened 2021-01-12
OKAPI-947 revert changes to _tenantPermissions

craigmcnally:
## Purpose
After the initial imlementation of OKAPI-947 was merged we realized that the breaking changes made to the _tenantPermissions interface weren't necessary and should be backed out.

Specifically:
* No need to provide separate `moduleFromId`
  * As a result we cam change `moduleToId` back to just `moduleId`.
* No need to provide `fromPerms`.  This information can be obtained by mod-permissions from querying the database.
  * As a result we can change `toPerms` back to just `perms`.
* Since breaking changes aren't required to the _tenantPermissions interface, we can remove tests which exercise the new (now reverted) v2.0 and backwards compatibility with v1.1 which is now the latest/current version of the interface again.

The changes to the module descriptor `renamedFrom` ARE still needed and are left alone.

See https://github.com/folio-org/okapi/pull/1057 for the original PR.


julianladisch (approved):

adamdickmeiss (approved):

hjiebsco (commented):
Looks like line 4727 in ModuleTest.java still uses "2.0". Is that intentional? 
`final String locHdrModule = createHeaderModule("2.0");`

craigmcnally:
> Looks like line 4727 in ModuleTest.java still uses "2.0". Is that intentional?
> `final String locHdrModule = createHeaderModule("2.0");`

Good catch @hjiebsco!  It's line 2747 btw :)

PR #1068, opened 2021-01-13
New document, "Disabling and undeploying unwanted FOLIO modules"

MikeTaylor:
This is based on hints from @wafschneider gave me yesterday in a Slack conversation. Hopefully it should make it possible for me to do Actual Work without my system going into paging frenzy.

### DIFF
--- a/doc/unwanted-modules.md
+++ b/doc/unwanted-modules.md
@@ -25,12 +25,12 @@ You can find these at http://localhost:9130/_/proxy/tenants/diku/modules
 
 (At the time of writing, there are exactly 100 of them!)
 
-Ignore the `names`, and consider the `id`s, discarding the version number. For example, if the output includes `edge-ncip-1.7.0-SNAPSHOT.17`, then you are interested in the ID `edge-ncip`.
+Ignore the `names`, and consider the `id`s, discarding the version number. For example, if the output includes `mod-email-1.10.0-SNAPSHOT.49`, then you are interested in the ID `mod-email`.
 
 
 ## 2. Choose the ones that are not wanted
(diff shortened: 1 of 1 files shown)


doc/unwanted-modules.md
```diff
@@ -0,0 +1,68 @@
...
+* [3. Check that disabling will not have catastrophic consequences](#3-check-that-disabling-will-not-have-catastrophic-consequences)
+* [4. Disable these modules (and those that depend on them)](#4-disable-these-modules-and-those-that-depend-on-them)
+* [5. Undeploy those modules including those that depend on them](#5-undeploy-those-modules-including-those-that-depend-on-them)
+
+In the development scenario, there is almost always only a single tenant in play, `diku`, so we will make that simplifying assumption in this document. To make the examples concrete, we will also assume that Okapi is running on localhost port 9130.
+
+
+## 1. Identify the modules running on the tenant of interest
+
+You can find these at http://localhost:9130/_/proxy/tenants/diku/modules
```
wafschneider, line 24: Those are modules _enabled_ for the tenant (and will include things that are less interesting, like UI modules). Actual _running_ modules are at `http://localhost:9130/_/discovery/modules`

wafschneider (approved):
Up to you about changing the endpoint to query, IMO it's a little clearer to query `/_/discovery/modules`, but I can see why `/_/proxy/tenant/<tenantId>/modules` is useful, too.

doc/unwanted-modules.md
MikeTaylor: Good point — now changed.

PR #1070, opened 2021-01-15
Fix illegal state exception check ended OKAPI-973

adamdickmeiss:
A lot of minor changes here, but the important part is that DockerModuleHandle.request now returns a HttpClientRequest and that both HttpClientRequest and HttpClientRespose is in the same scope!

This PR also has tweaks to make it work better on Windows (at least as far as testing is concerned)


okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
```diff
@@ -70,6 +70,21 @@ public ProcessModuleHandle(Vertx vertx, LaunchDescriptor desc, String id,
...
+    return c.connect(port, "localhost")
+        .compose(socket ->
+                socket.close().compose(y -> {
+                  if (iter == 0) {
+                    return Future.failedFuture(
+                        messages.getMessage("11502", Integer.toString(port)));
+                  }
+                  Promise<Void> promise = Promise.promise();
+                  vertx.setTimer(100, x -> waitPortOpen(c, iter - 1)
+                      .onComplete(promise::handle));
```
julianladisch, line 83:
```suggestion
                  vertx.setTimer(100, x -> waitPortOpen(c, iter - 1).onComplete(promise::handle));
```
Avoid the misleading line break.

okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
```diff
@@ -70,6 +70,21 @@ public ProcessModuleHandle(Vertx vertx, LaunchDescriptor desc, String id,
...
+                  return promise.future();
+                }), x  -> Future.succeededFuture());
```
julianladisch, line 85: `x` should be renamed to `noSocker` and vertically aligned with `socket` to better show that it is the second argument of the two-argument compose method.

okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
adamdickmeiss: OK. Allow me to put waitPortOpen on its own line instead.

okapi-core/src/main/java/org/folio/okapi/service/impl/ProcessModuleHandle.java
adamdickmeiss: Thx, much better.

okapi-core/src/test/java/org/folio/okapi/PullTest.java
```diff
@@ -115,6 +118,13 @@ private void td(TestContext context, Async async) {
...
+    if (otherServer != null) {
+      otherServer.close().onComplete(res -> {
+        otherServer = null;
+        td(context, async);
+      });
+      return;
+    }
     vertx.close(context.asyncAssertSuccess());
     async.complete();
   }
```
julianladisch, line 130: [Vertx.close](https://vertx.io/docs/apidocs/io/vertx/core/Vertx.html#close--) automatically closes all resources (Verticles, servers, ...) opened by it, no need to do it manually.
Do you have a use case or reproducer where this doesn't work? If yes this is a bug in Vertx and should be reported upstream.

    @After
    public void tearDown(TestContext context) {
      vertx.close().onComplete(context.asyncAssertSuccess());
    }

or

    @After
    public void tearDown(TestContext context) {
      vertx.close(context.asyncAssertSuccess());
    }

okapi-core/src/test/java/org/folio/okapi/PullTest.java
```diff
@@ -50,17 +52,18 @@ private void setupOtherHttpServer(TestContext context, Async async) {
...
+        .requestHandler(router)
+        .listen(
+            port3,
+            result -> {
+              if (result.failed()) {
+                context.fail(result.cause());
+              }
+              async.complete();
+            }
+        );
```
julianladisch, line 66: Use futurisation to avoid all the manual error handling:

    @Before
    public void setUp(TestContext context) {
      logger.debug("staring PullTest");
      vertx = Vertx.vertx();

      DeploymentOptions opt1 = new DeploymentOptions()
          .setConfig(new JsonObject().put("port", Integer.toString(port2)));
      DeploymentOptions opt2 = new DeploymentOptions()
          .setConfig(new JsonObject().put("port", Integer.toString(port1)));
      HttpServerOptions so = new HttpServerOptions().setHandle100ContinueAutomatically(true);

      vertx.deployVerticle(MainVerticle.class, opt1)
          .compose(x -> vertx.deployVerticle(MainVerticle.class, opt2))
          .compose(x -> vertx.createHttpServer(so)
              .requestHandler(Router.router(vertx))
              .listen(port3))
          .onComplete(context.asyncAssertSuccess());
    }

okapi-core/src/test/java/org/folio/okapi/PullTest.java
adamdickmeiss: ok.. I was probably being paranoid.. I'll test a little on my slow Windows (because in Vmware player) .

julianladisch (approved):
Looks good, thanks!

PR #1072, opened 2021-01-18
Default for RestAssured log().ifValidationFails()

adamdickmeiss:
A minor refactor to this, otherwise, old test.


julianladisch (approved):
Good default, thanks!

okapi-core/src/test/java/org/folio/okapi/ModuleTest.java
```diff
@@ -102,6 +102,7 @@ private void assertEmptyReport(RestAssuredClient c) {
   @BeforeClass
   public static void setUpBeforeClass() {
     api = RamlLoaders.fromFile("src/main/raml").load("okapi.raml");
+    RestAssured.enableLoggingOfRequestAndResponseIfValidationFails();
```
hjiebsco, line 105: Nice.

hjiebsco (approved):

PR #1073, opened 2021-01-18
Make permissionName a required property OKAPI-974

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/bean/ModuleDescriptor.java
```diff
@@ -192,6 +192,7 @@ public void setProvidedHandler(String id, String version, RoutingEntry... entrie
     interfaceDescriptor.setId(id);
     interfaceDescriptor.setVersion(version);
     interfaceDescriptor.setHandlers(entries);
+    setProvides(interfaceDescriptors);
```
hjiebsco, line 195: Good catch

okapi-core/src/main/java/org/folio/okapi/bean/ModuleDescriptor.java
```diff
@@ -294,8 +295,21 @@ public void setLaunchDescriptor(LaunchDescriptor launchDescriptor) {
...
   public void setPermissionSets(Permission[] permissionSets) {
     this.permissionSets = permissionSets;
+    if (permissionSets != null) {
+      for (Permission permission : permissionSets) {
+        String permissionName = permission.getPermissionName();
+        if (permissionName == null) {
+          throw new IllegalArgumentException("Missing permissionName");
+        }
+      }
+    }
```
hjiebsco, line 312: Maybe move these before Line 304? It won't make a difference here, but validation before assign seems more natural.

okapi-core/src/test/java/org/folio/okapi/bean/BeanTest.java
```diff
@@ -286,21 +238,13 @@ public void testModuleDescriptorTimers() {
...
-        ModuleDescriptor.class);
-      String pretty = Json.encodePrettily(md);
-    } catch (DecodeException ex) {
-      ex.printStackTrace();
-      fail = 400;
-    }
-    assertEquals(0, fail);
+    ModuleDescriptor moduleDescriptor = Json.decodeValue(docModuleDescriptor, ModuleDescriptor.class);
+    moduleDescriptor.setPermissionSets(null);
+    assertNull(moduleDescriptor.getPermissionSets());
```
hjiebsco, line 243: Do not see the value of these two lines.

hjiebsco (approved):

okapi-core/src/test/java/org/folio/okapi/bean/BeanTest.java
adamdickmeiss: Makes coverage 100%, rather than 91%

okapi-core/src/main/java/org/folio/okapi/bean/ModuleDescriptor.java
adamdickmeiss: ok

okapi-core/src/main/java/org/folio/okapi/bean/ModuleDescriptor.java
adamdickmeiss: My editor informed me (IntelliJ IDEA)

PR #1078, opened 2021-02-02
Dump docker response when decoding JSON fails

adamdickmeiss:
This is a PR which simply makes the Docker Handling of Okapi dump the response body when decoding it as JSON fails.

This is to investigate https://issues.folio.org/browse/OKAPI-978


adamdickmeiss:
No need to use this for b4.6 as this has been reproduced locally.

PR #1079, opened 2021-02-02
Fix exception on docker pull OKAPI-978

adamdickmeiss:
Docker does not return JSON. It returns lines of JSON objects.
We just consider the first one which is all we need.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
+++ b/okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
@@ -268,7 +268,7 @@ public class DockerModuleHandle implements ModuleHandle {
       return getUrl("/images/" + image + "/json", "getImage");
     }
     if (dockerRegistries.isEmpty()) {
-      logger.warn(DOCKER_REGISTRIES_EMPTY_LIST);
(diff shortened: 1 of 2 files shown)


okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
```diff
@@ -229,11 +229,19 @@ private void logHandler(Buffer b) {
...
+      if (idx != -1) {
+        line = line.substring(0, idx);
+      }
       try {
-        JsonObject b = body.toJsonObject();
+        JsonObject b = new JsonObject(line);
         promise.complete(b);
       } catch (DecodeException e) {
         logger.warn("{}", e.getMessage(), e);
+        logger.warn("while decoding {}", line);
```
julianladisch, line 244:
```suggestion
        logger.warn("{} while decoding {}", e.getMessage(), line);
```

julianladisch (approved):
Looks good, thanks!

okapi-core/src/main/java/org/folio/okapi/service/impl/DockerModuleHandle.java
adamdickmeiss: oh. forgot that one.. But the message was also shown previous line.

PR #1080, opened 2021-02-03
Extend okapi-curl with login facility OKAPI-980

adamdickmeiss:
Using "okapi-curl login" will cause okapi-curl to interactively
ask for tenant, username and password.
The normal call will no longer prepend /. It will also not pass
X-Okapi-Token as this is redundant (part of token).

### DIFF
--- a/util/okapi-curl
+++ b/util/okapi-curl
@@ -50,7 +50,7 @@ fi
 path="$1"
 shift
 
-exec curl \
+exec curl -w '\n' \
 	-H "X-Okapi-Token:$OKAPI_TOKEN" \
 	-H "Content-Type:application/json" \
 	-H "Accept:*/*" \
--- a/util/okapi-curl
+++ b/util/okapi-curl
@@ -3,26 +3,56 @@
 # Edit the `.okapi` file in your home directory to contain something like:
 #
 #	OKAPI_URL=https://east-okapi.folio-dev.indexdata.com
-#	OKAPI_TENANT=reshare_east
 #	OKAPI_TOKEN=xxxxxxxx
+#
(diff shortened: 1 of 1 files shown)


util/okapi-curl
```diff
@@ -3,26 +3,56 @@
 # Edit the `.okapi` file in your home directory to contain something like:
 #
 #	OKAPI_URL=https://east-okapi.folio-dev.indexdata.com
-#	OKAPI_TENANT=reshare_east
```
MikeTaylor, line 6: I think here you meant to delete the TOKEN line, not the TENANT line — is that right?

util/okapi-curl
```diff
@@ -3,26 +3,56 @@
...
+fi
+
+. ~/.okapi
+if test -z "$OKAPI_URL"; then
+	echo "OKAPI_URL must be given in .okapi"
+	exit 1
+fi
+if test "$1" = "login"; then
+	echo -n "Tenant:"
+	read tenant
```
MikeTaylor, line 30: To me, prompting for user/password is obviously correct, but tenant is part of the identity of the service that you're logging into, so it should be already in the `~/.okapi` file. The process is
1. Set up URL and tenant in that file
2. Log in to the service that they define.

MikeTaylor (changes requested):
This is a very practical addition, thank you. Let's just sort out the tenant handling and we're good.

util/okapi-curl
adamdickmeiss: No.

util/okapi-curl
adamdickmeiss: No. You supply tenant, user and password. Rest is in token.

PR #1082, opened 2021-02-06
Inform mod-permissions when module is disabled OKAPI-982

adamdickmeiss:
1. Removing module sends module with empty list via _tenantPermission so that permissions module may mark those permissions as no longer in use by the module.
2. Config option: `enable_system_auth` to control whether to call auth even if it's enabled. Highly insecure, so this is something that can not be controlled at runtime.
3. install parameter `invoke` is a string rather than a `boolean`. Values `true` and `false` honored as before. Other value is regular expression for controlling which modules to invoke.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
@@ -673,11 +673,9 @@ public class TenantManager implements Liveness {
       case "2.0":
         if (mdTo != null) {
           pl = new PermissionList(mdTo.getId(), mdTo.getExpandedPermissionSets());
-        } else if (mdFrom != null) {
+        } else  {
           // attempt an empty list for the module being disabled
           pl = new PermissionList(mdFrom.getId(), new Permission[0]);
-        } else {
-          return Future.succeededFuture();
         }
         break;
       default:
--- a/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
@@ -340,8 +340,7 @@ public class TenantManager implements Liveness {
   private Future<Void> invokeTenantInterface(Tenant tenant, TenantInstallOptions options,
                                              ModuleDescriptor mdFrom, ModuleDescriptor mdTo,
                                              ProxyContext pc) {
-
-    if (!options.getInvoke()) {
+    if (!options.checkInvoke(mdTo.getId())) {
       return Future.succeededFuture();
     }
     JsonObject jo = new JsonObject();
@@ -405,7 +404,7 @@ public class TenantManager implements Liveness {
    */
   private Future<Void> invokePermissions(Tenant tenant, TenantInstallOptions options,
       ModuleDescriptor mdFrom, ModuleDescriptor mdTo, ProxyContext pc) {
-    if (!options.getInvoke()) {
+    if (!options.checkInvoke(mdTo.getId())) {
       return Future.succeededFuture();
     }
     if (mdTo != null && mdTo.getSystemInterface("_tenantPermissions") != null) {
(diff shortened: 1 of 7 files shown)


okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
```diff
@@ -69,12 +69,23 @@ public boolean getDepCheck() {
...
 
-  public boolean getInvoke() {
-    return invoke;
+  /**
+   * Check if module is to be invoked during install.
+   * @param id module ID.
+   * @return whether to invoke the module.
+   */
+  public boolean checkInvoke(String id) {
+    if (invoke == null || "true".equals(invoke)) {
```
craigmcnally, line 82: Do we want to try and handle "True" / "TRUE" / etc. as well?  `Boolean.parseBoolean(invoke)` might be a better check here?

okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
```diff
@@ -69,12 +69,23 @@ public boolean getDepCheck() {
...
+      return true;
+    }
+    if ("false".equals(invoke)) {
```
craigmcnally, line 85: Likewise, do we want to try and handle False / FALSE /etc. here?  maybe via something like `"false".equalsIgnoreCase(invoke)`

okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
craigmcnally: or as suggested below use `String.equalsIgnoreCase(...)` ?

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -651,31 +650,41 @@ private void fireTimer(Tenant tenant, ModuleDescriptor md, RoutingEntry re, Stri
...
+          pl = new PermissionList(mdFrom.getId(), new Permission[0]);
+        }
         break;
       default:
         return Future.failedFuture(new OkapiError(ErrorType.USER,
             "Unknown version of _tenantPermissions interface in use " + permIntVer + "."));
     }
     String pljson = Json.encodePrettily(pl);
-    logger.debug("tenantPerms Req: {}", pljson);
+    logger.info("tenantPerms Req: {}", pljson);
```
craigmcnally, line 687:
```suggestion
    logger.info("_tenantPermissions Req: {}", pljson);
```

craigmcnally (approved):
Looks good - just a few minor suggestions - You can take them or leave them.

okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
adamdickmeiss: The code elsewhere is pretty strict with this .. In ModuleUtil.getParmBoolean where anything but `true` or `false` will be reported as error. So I think we should make it strict to keep the style.

okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
adamdickmeiss: :-)

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: fine

okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
adamdickmeiss: Someday, someone will make a `true`module.

PR #1084, opened 2021-02-12
Use HTTP check for module readiness in deployment OKAPI-984

adamdickmeiss:
Use WebClient on module host+port to see if module is alive, rather
than just TCP socket agsinst host and port. The latter is unreliable
in cases of Containers where exposed port is ready "immediately". This
in turn fixes Okapi startup with mod-permissions being called for
the purpose of upgrading the okapi module for a tenant.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/util/TcpPortWaiting.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/TcpPortWaiting.java
@@ -4,8 +4,8 @@ import com.zaxxer.nuprocess.NuProcess;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
-import io.vertx.core.net.NetClient;
-import io.vertx.core.net.NetClientOptions;
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.ext.web.client.WebClientOptions;
 import org.apache.logging.log4j.Logger;
 import org.folio.okapi.common.Messages;
 import org.folio.okapi.common.OkapiLogger;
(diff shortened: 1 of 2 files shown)


okapi-core/src/main/java/org/folio/okapi/util/TcpPortWaiting.java
```diff
@@ -34,24 +34,26 @@ public TcpPortWaiting(Vertx vertx, String host, int port) {
   }
 
   private Future<Void> tryConnect(NuProcess process, int count) {
-    NetClientOptions options = new NetClientOptions().setConnectTimeout(MILLISECONDS);
-    NetClient c = vertx.createNetClient(options);
+    WebClientOptions options = new WebClientOptions().setConnectTimeout(MILLISECONDS);
```
julianladisch, line 37:
```suggestion
    // don't use NetClient because container ports are immediately ready, instead check for HTTP response
    WebClientOptions options = new WebClientOptions().setConnectTimeout(MILLISECONDS);
```

okapi-core/src/main/java/org/folio/okapi/util/TcpPortWaiting.java
```diff
@@ -34,24 +34,27 @@ public TcpPortWaiting(Vertx vertx, String host, int port) {
...
+                id -> tryConnect(process, count + 1).onComplete(promise));
+          } else {
             promise.fail(messages.getMessage("11501",
-                Integer.toString(port), res.cause().getMessage()))
-        );
-      }
-    });
+                Integer.toString(port), cause.getMessage()));
+          }
+        });
```
julianladisch, line 57: The WebClient stuff should be moved into a separate method, and Future methods should be used:

```
  private Future<Void> tryConnect(NuProcess process, int count) {
    logger.info("tryConnect() host {} port {} count {}", host, port, count);
    return tryConnect()
        .onSuccess(res -> {
          logger.info("Connected to service at host {} port {} count {}", host, port, count);
        })
        .recover(cause -> {
          if (count < maxIterations && (process == null || process.isRunning())) {
            return Future.future(promise ->
              vertx.setTimer((long) (count + 1) * MILLISECONDS,
                id -> tryConnect(process, count + 1).onComplete(promise)));
          } else {
            return Future.failedFuture(messages.getMessage("11501",
                Integer.toString(port), cause.getMessage()));
          }
        });
  }

  private Future<Void> tryConnect() {
    // don't use NetClient because container ports are immediately ready, instead check for HTTP
    WebClientOptions options = new WebClientOptions().setConnectTimeout(MILLISECONDS);
    WebClient c = WebClient.create(vertx, options);
    return c.get(port, host, "/")
        .send()
        .<Void>mapEmpty()
        .onComplete(result -> c.close());
  }
```

adamdickmeiss: More lines, but neat that the WebClient usage is isolated.

julianladisch (approved):
Thanks, looks good!

PR #1090, opened 2021-02-25
Fix modules require does not consider optional deps OKAPI-987

adamdickmeiss:
Steal UtilityClassTester from RMB.


hjiebsco (approved):
Is `UtilityClassTester` from RMB?

adamdickmeiss:
> Is `UtilityClassTester` from RMB?

Yes. This was added by @julianladisch May 2017.

PR #1092, opened 2021-02-26
Config deploy.waitIterations OKAPI-990

adamdickmeiss:


okapi-common/src/main/java/org/folio/okapi/common/Config.java
```diff
@@ -68,4 +68,19 @@ public static Boolean getSysConfBoolean(String key, Boolean def, JsonObject conf
...
+   * @param def default value (may be null).
+   * @param conf JSON object configuration.
+   * @return value (possibly null).
+   */
+  public static Integer getSysConfInteger(String key, Integer def, JsonObject conf) {
+    final String v = System.getProperty(key);
+    if (v == null || v.isEmpty()) {
+      return conf.getInteger(key, def);
+    }
+    return Integer.parseInt(v);
```
hjiebsco, line 84: Any concern if `v` cannot be parsed to integer?

hjiebsco (approved):
Left a minor comment about `parseInt`. BTW how did you come up `x*(x+1) * 0.2 seconds`? :)

okapi-common/src/main/java/org/folio/okapi/common/Config.java
adamdickmeiss: I'll check and, perhaps, make a test.

adamdickmeiss:
> Left a minor comment about `parseInt`. BTW how did you come up `x*(x+1) * 0.2 seconds`? :)

The deployment waits 0.2 * (1 + 2 + 3 + 4 + ..n) = 0.2 * (n+1) * n  .. It is due to the fact that it will poll slower and slower.. Which IMHO makes sense.. If things is slow, it could be a lot slower...

okapi-common/src/main/java/org/folio/okapi/common/Config.java
adamdickmeiss: I tried to the property to "foo" and Okapi will fail the deployment and return a nice 400. And the log will include the backtrace ok. So I think we're good to go!

doc/guide.md
```diff
@@ -2816,6 +2816,9 @@ leave it unmodified (default).
   Defaults to 10000.
 * `token_cache_ttl_ms`: Time to live in milliseconds for token cache entries.
   Defaults to 180000 (3 minutes).
+* `deploy.waitIterations`: Maximum number of iterations for deployment - before
```
julianladisch, line 2819: Environment variables with periods/dots in their names should be avoided because a period is not POSIX compliant and therefore some shells, notably, the BusyBox /bin/sh included in Alpine Linux, strip them.
References:
* warning in OpenJDK docs: https://hub.docker.com/_/openjdk/
* POSIX.1-2017: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html

doc/guide.md
```diff
@@ -2816,6 +2816,9 @@ leave it unmodified (default).
...
+  giving up (readiness check). A value, `n`, corresponds to roughly  `n*n*0.2` seconds.
+  This value, if set, overrides the `waitIterations` in the launch descriptor.
```
julianladisch, line 2821: Can you mention the default?

julianladisch, line 2821: Can you extend
https://github.com/folio-org/okapi/blob/master/dist/okapi.conf
https://github.com/folio-org/okapi/blob/master/dist/okapi.sh
accordingly?

okapi-core/src/main/java/org/folio/okapi/util/TcpPortWaiting.java
```diff
@@ -3,6 +3,7 @@
 import com.zaxxer.nuprocess.NuProcess;
 import io.vertx.core.Future;
 import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonObject;
```
julianladisch, line 6: Can you remove this unused import?

doc/guide.md
adamdickmeiss: Not an environment variable. This section in the guide lists Java properties.

doc/guide.md
adamdickmeiss: The okapi.conf and okapi.sh are very problematic. Any time we need to add *anything* they have to be updated. Had they just made an option to refer to user-defined okapi.conf (with Vert.x properties) once and for all, NOTHING would have to be updated.

It's probably legacy from times when okapi.conf did not include "everything".

PR #1095, opened 2021-02-27
OKAPI-991: Fix deploy_waitIterations (was deploy.waitIterations)

julianladisch:
* Fix the option name by renaming from deploy.waitIterations to deploy_waitIterations.
* Add the option to okapi.conf and okapi.sh
* Minor issues raised during the OKAPI-990 pull request https://github.com/folio-org/okapi/pull/1092

Environment variables with periods/dots in their names should be
avoided because a period is not POSIX compliant and therefore some
shells, notably, the BusyBox /bin/sh included in Alpine Linux, strip them.
References:
* warning in OpenJDK docs: https://hub.docker.com/_/openjdk/
* POSIX.1-2017: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html


hjiebsco (approved):
I am fine with change but I will leave @adamdickmeiss to decide if it is urgent enough to make a patch release. Also should we consider refactoring `http.port` to `http_port`?

NEWS.md
```diff
@@ -1,11 +1,15 @@
+## 4.7.1
+
+ * [OKAPI-991](https://issues.folio.org/browse/OKAPI-991) Fix deploy\_waitIterations (was deploy.waitIterations)
+
 ## 4.7.0 2021-02-26
```
adamdickmeiss, line 5: No news in PRs except when a release is made.

adamdickmeiss (changes requested):
I don't like dots in the config name. But why did I keep it? To keep compatibility. And even more so: this is *not* an environment varilable. This is a Vert.x config property or Java property which do not impose such limitations.

Not a bug, but if you really think the name is a problem we should keep the existing config recognized. No need to make a breaking change.

adamdickmeiss:
> I am fine with change but I will leave @adamdickmeiss to decide if it is urgent enough to make a patch release. Also should we consider refactoring `http.port` to `http_port`?

Also here.. I think we are confusing Java properties with environment variables. So http.port should be fine. And like the other is never inspected as an environment variable.

hjiebsco:
> > I am fine with change but I will leave @adamdickmeiss to decide if it is urgent enough to make a patch release. Also should we consider refactoring `http.port` to `http_port`?
> 
> Also here.. I think we are confusing Java properties with environment variables. So http.port should be fine. And like the other is never inspected as an environment variable.

Got it. @adamdickmeiss. Thanks for clarifying it.

hjiebsco (commented):
Maybe consider to make the new config a constant to avoid repeating it in few places.

adamdickmeiss:
> Maybe consider to make the new config a constant to avoid repeating it in few places.

It's only mentioned twice.. But perhaps we should make a utility class with *all* observed properties.. so to have them listed, in one place.

julianladisch:
Please explain why reading deploy.waitIterations from okapi.conf in https://github.com/folio-org/okapi/blob/v4.7.1/dist/okapi.sh#L10 isn't an environment variable.

I cannot find any usage of deploy.waitIterations outside of Okapi: https://github.com/search?q=org%3Afolio-org+%22deploy+waitIterations%22&type=code

And the deploy.waitIterations was an undocumented feature before OKAPI-990.

> should we consider refactoring `http.port` to `http_port`?

In okapi.conf it has already been refactored: https://github.com/folio-org/okapi/blob/v4.7.1/dist/okapi.sh#L135-L148

adamdickmeiss:
> Please explain why reading deploy.waitIterations from okapi.conf in https://github.com/folio-org/okapi/blob/v4.7.1/dist/okapi.sh#L10 isn't an environment variable.

It is not mentioned there. If people want to change it, they can in existing Okapi do 
```
OKAPI_JAVA_OPTS += "-D deploy.waitIterations=100"
```

> 
> I cannot find any usage of deploy.waitIterations outside of Okapi: https://github.com/search?q=org%3Afolio-org+%22deploy+waitIterations%22&type=code

OK? But why change it? It was a contribution from normal channels who made the PR .. Obviously he/she is using it (China, I think).

> 
> And the deploy.waitIterations was an undocumented feature before OKAPI-990.
Yes.
> 
> > should we consider refactoring `http.port` to `http_port`?
> 
> In okapi.conf it has already been refactored: https://github.com/folio-org/okapi/blob/v4.7.1/dist/okapi.sh#L135-L148

Right.

In short, there's no reason to change it. If you think it is important enough for people to have a setting in okapi.conf to set it, then I'm ready to accept that. I would leave to to devops guys to worry about. At *this time* they can just edit /etc/default/okapi .

adamdickmeiss (changes requested):
So Okapi depends on RMB and RMB depends on Okapi. Please not go there.

okapi-core/pom.xml
```diff
@@ -168,6 +168,11 @@
       <version>4.0.3</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.folio</groupId>
+      <artifactId>testing</artifactId>
+      <scope>test</scope>
+    </dependency>
```
julianladisch, line 175: org.folio:testing depends on junit:junit only: https://github.com/folio-org/raml-module-builder/blob/v32.2.0/testing/pom.xml

We cannot move UtilityClassTester.java from org.folio:testing to org.folio:folio-service-tools-test
https://github.com/folio-org/folio-service-tools/blob/v1.6.1/folio-service-tools-test/pom.xml
because org.folio:folio-service-tools-test depends on org.folio:domain-models-runtime.

adamdickmeiss: It's possible to make a okapi testing pom and then RMB might use that. But it doesn't have to , immediately. Then the direction is one way.

julianladisch: https://issues.folio.org/browse/FOLIO-3052 "Move org.folio:testing into own repository"

julianladisch:
> Obviously he/she is using it

A single person is using a non-documented feature. For me this is not enough to have inconsistent names (underscore versus dot) in okapi.conf (`deploy_waitIterations=`) and on the command line (`-Ddeploy.waitIterations=`).

okapi-core/pom.xml
adamdickmeiss: > > Obviously he/she is using it
> 
> A single person is using a non-documented feature. For me this is not enough to have inconsistent names (underscore versus dot) in okapi.conf (`deploy_waitIterations=`) and on the command line (`-Ddeploy.waitIterations=`).

Remind me again , why `http.port` was seen as better naming than just `port`? We have applied that in a number of places already.

I'm not saying that dots are better than underscores, but I made a judgement call that keeping behavior was better than changing. That's the contract we make with semantic versioning.

If you feel strongly about it (which I don't understand), I suggest you make a PR that accepts both.

I wouldn't spend time on it. There are more important tasks out there.

adamdickmeiss (approved):

PR #1096, opened 2021-03-01
OKAPI-960 Add PERSONAL_DATA_DISCLOSURE form

hjiebsco:
Please see https://issues.folio.org/browse/OKAPI-960 Add personal data disclosure form 

Note:
- okapi-common logging uses `userId`
- `IP` is used in https://github.com/folio-org/okapi/blob/master/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java#L584

2021-03-03
Add below item as personal data per core platform discussion

- User search queries

### DIFF
--- a/PERSONAL_DATA_DISCLOSURE.md
+++ b/PERSONAL_DATA_DISCLOSURE.md
@@ -1,0 +1,58 @@
+
+
+## Overview
+The purpose of this form is to disclose the types of personal data stored by each module.  This information enables those hosting FOLIO to better manage and comply with various privacy laws and restrictions, e.g. GDPR.
+
+It's important to note that personal data is not limited to that which can be used to identify a person on it's own (e.g. Social security number), but also data used in conjunction with other data to identify a person (e.g. date of birth + city + gender).
+
+For the purposes of this form, "store" includes the following:
+* Persisting to storage - Either internal (e.g. Postgres) or external (e.g. S3, etc.) to FOLIO
+* Caching - In-memory, etc.
+* Logging
+* Sending to an external piece of infrastructure such as a queue (e.g. Kafka), search engine (e.g. Elasticsearch), distributed table, etc.
+
+## Personal Data Stored by This Module
+- [ ] This module does not store any personal data.
+- [ ] This module provides [custom fields](https://github.com/folio-org/folio-custom-fields).
+- [ ] This module stores fields with free-form text (tags, notes, descriptions, etc.)
+- [ ] This module caches personal data
+---
+- [ ] First name
+- [ ] Last name
+- [ ] Middle name
+- [x] Pseudonym / Alias / Nickname / Username / User ID
+- [ ] Gender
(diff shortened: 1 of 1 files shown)


adamdickmeiss (commented):
Can this be in doc instead and the README.md have a pointer to it?
I like it when the number of files in the root directory is minimal.

adamdickmeiss:
> Can this be in doc instead and the README.md have a pointer to it?
> I like it when the number of files in the root directory is minimal.

Guess it's been decided that this must live in top-level. Sigh.

PERSONAL_DATA_DISCLOSURE.md
```diff
@@ -0,0 +1,58 @@
...
+- [ ] Racial or ethnic origin
+- [ ] Address
+- [ ] Location information
+- [ ] Phone numbers
+- [ ] Passport number / National identification numbers
+- [ ] Driver’s license number
+- [ ] Social security number
+- [ ] Email address
+- [ ] Web cookies
+- [x] IP address
```
adamdickmeiss, line 36: Okapi does not log the IP-address of the originating user or even the preceding proxy, AFAIK:

adamdickmeiss (approved):

PERSONAL_DATA_DISCLOSURE.md
```diff
@@ -0,0 +1,58 @@
...
+- [ ] Geolocation data
+- [ ] Financial information
+- [ ] Logic or algorithms used to build a user/profile
+<!--- - [ ] Other personal data - Please list as needed -->
+<!--- - [ ] Other personal data - Please list as needed -->
```
julianladisch, line 41: Okapi logs date, time and all queries a user submits. Example:
```
2021-03-02T12:44:56,486 INFO  ProxyContext         102758/inventory RES 200 377212us mod-inventory-16.1.3 http://172.17.0.1:9157/inventory/instances?limit=100&query=%28keyword%20all%20%22foo%22%29%20sortby%20title
```

hjiebsco: Are those considered as personal data as well? @craigmcnally, can you confirm?

PERSONAL_DATA_DISCLOSURE.md
hjiebsco: Depends on how you set up FOLIO. The request/client IP could be from originating user, right?

adamdickmeiss: Sure, very unlikely though. If *I* were to track users, I'd use the request-Id.. :-)

PERSONAL_DATA_DISCLOSURE.md
hjiebsco: Per core platform standup discussion, `date` and `time` are not personal data. I added `user search queries` as a separate personal data item. Yes, it could be used to build user profile but Okapi currently does not have logic or algorithms to do so.

PR #1101, opened 2021-03-08
Reload permissions immediately on enable (not upgrade) OKAPI-996

adamdickmeiss:
1. reload permissions when mod-permissions is being enabled (not upgraded)
2. reload permissions at end of install when mod-permissions is upgraded.

Previously permissions was reloaded at the end always.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
@@ -275,25 +275,19 @@ public class TenantManager implements Liveness {
               ? enableAndDisableCheck(tenant, mdFrom.result(), mdTo.result())
               : Future.succeededFuture())
         .compose(x -> enableAndDisableModule(tenant, options, mdFrom.result(),
-            mdTo.result(), true, pc));
+            mdTo.result(), pc));
   }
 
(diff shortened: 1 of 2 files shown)


okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -447,21 +441,23 @@ private void waitTenantInit(Tenant tenant, ModuleInstance getInstance,
    * Announce permissions for a set of modules to a permissions module.
```
craigmcnally, line 441:
```suggestion
   * Conditionally announce permissions for a set of modules to a permissions module.  This will only happen if the permissions module is being enabled for the first time, not when it's upgraded.
```

hjiebsco (approved):
LGTM. Keep fingers crossed :)

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: I'll add it.. Just realize that it may be called on last refresh and fooled by mdFrom = null (you probably saw that already)

craigmcnally (approved):

PR #1106, opened 2021-03-12
okapi-testing with UtilityClassTester OKAPI-999

adamdickmeiss:


julianladisch (approved):
Looks good, thanks!

okapi-testing/src/main/java/org/folio/okapi/testing/UtilityClassTester.java
```diff
@@ -18,15 +18,15 @@ private UtilityClassTester() {
...
-                       This is save because
-                       - it runs during unit tests only, and
-                       - it invokes the constructor only, no write access, no i/o.
-                       There is no other way to test whether a private constructor
-                       throws an Exception.
-                     */
+      "squid:S1166",  /* suppress "Either log or rethrow this exception" */
+      "squid:S3011",  /* suppress "Changing accessibility is security sensitive" and
+                         suppress "Make sure that this accessibility update is safe here."
+                         This is save because
```
hjiebsco, line 24: typo `save`

okapi-testing/src/test/java/org/folio/okapi/testing/UtilityClassTesterTest.java
```diff
@@ -0,0 +1,144 @@
...
+
+import org.junit.Test;
+
+public class UtilityClassTesterTest {
+  @Test
+  public void selfTest() {
+    UtilityClassTester.assertUtilityClass(UtilityClassTester.class);
+  }
+
+  public void assertMessage(Class<?> clazz, String message) {
```
hjiebsco, line 15: Need to be `public`?

okapi-testing/src/main/java/org/folio/okapi/testing/UtilityClassTester.java
adamdickmeiss: yep

okapi-testing/src/test/java/org/folio/okapi/testing/UtilityClassTesterTest.java
adamdickmeiss: private is fine.

PR #1107, opened 2021-03-15
Schedule facility (ala cron) OKAPI-1000

adamdickmeiss:
routing entry has "schedule" property as an alternative to
"unit" and "duration". The schedule object at this stage takes a cron-property that is a crontab entry.

PR #1108, opened 2021-03-16
Cron scheduler using cron utils OKAPI-1000

adamdickmeiss:
But it doesn't work. At the point of an event, the
executionTime.timeToNextExecution will return repeated 0 values.
The home-brewn Scheduler will not. It will at the point of fire
move the minute up by 1, and thus, hit next event.


adamdickmeiss:
Closing as this is done in #1107 (using cron-utils)

PR #1109, opened 2021-03-22
Pull: ignore MDs with unsupported features OKAPI-1001

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/managers/PullManager.java
```diff
@@ -127,8 +129,17 @@ public PullManager(Vertx vertx, ModuleManager moduleManager) {
                     promise.fail(new OkapiError(ErrorType.USER, body.toString()));
                     return;
                   }
-                  ModuleDescriptor[] ml = Json.decodeValue(body.toString(),
-                      ModuleDescriptor[].class);
+                  List<ModuleDescriptor> ml = new LinkedList<>();
+                  JsonArray objects = body.toJsonArray();
+                  for (int pos = 0; pos < objects.size(); pos++) {
+                    JsonObject mdObj = objects.getJsonObject(pos);
+                    String id = mdObj.getString("id");
```
hjiebsco, line 136: This can be moved to `catch` block since it is not used in normal cases.

hjiebsco (approved):

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -108,7 +109,8 @@ public ModuleManager(ModuleStore moduleStore, boolean local) {
         }
       }
       if (check) {
-        String res = DepResolution.checkDependencies(tempList.values(), newList);
+        String res = DepResolution.checkDependencies(tempList.values(), newList,
+            removeIfMissingDep);
         if (!res.isEmpty()) {
```
hjiebsco, line 114: If both `check` and `removeIfMissingDep` are true, the `res` will always be empty, correct?

hjiebsco (approved):

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
adamdickmeiss: Correct. No errors would be reported, then.

PR #1110, opened 2021-03-25
Timer management OKAPI-969

adamdickmeiss:


doc/guide.md
```diff
@@ -3128,6 +3130,46 @@ Examples:
...
+
+```
+curl http://localhost:9130/_/proxy/tenants/testlib/timers
+```
+
+The first entry fires a timer every 20 seconds. Disable it with:
+
+```
+curl -XPATCH -d'{"id":"0_test-timer","routingEntry":{"delay":"0"}}' \
+    http://localhost:9130/_/proxy/tenants/testlib/timers
```
hjiebsco, line 3170: Would it be possible to simplify the `PATCH` to use id? For example:
``` 
curl -XPATCH -d'{"delay":"0"}' \
    http://localhost:9130/_/proxy/tenants/testlib/timers/0_test-timer
```

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
```diff
@@ -337,8 +342,16 @@ public void start(Promise<Void> promise) {
   }
 
   private Future<Void> startRedeploy() {
-    return discoveryManager.restartModules()
-        .compose(res -> tenantManager.startTimers(discoveryManager, okapiVersion));
```
hjiebsco, line 341: Is the idea to maintain changed timers? If yes, is there an easy to reset to default MD timers?

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
```diff
@@ -330,6 +337,24 @@ public static ModuleDescriptor moduleDescriptor(String okapiVersion) {
...
+        + "    \"methods\" :  [ \"GET\" ],"
+        + "    \"pathPattern\" : \"/_/proxy/tenants/{tenantId}/timers\","
+        + "    \"permissionsRequired\" : [  \"okapi.proxy.tenants.timers.list\" ], "
+        + "    \"permissionsRequiredTenant\" : [ ], "
+        + "    \"type\" : \"internal\" "
+        + "   }, {"
+        + "    \"methods\" :  [ \"PATCH\" ],"
+        + "    \"pathPattern\" : \"/_/proxy/tenants/{tenantId}/timers\","
+        + "    \"permissionsRequired\" : [  \"okapi.proxy.tenants.timers.patch\" ], "
+        + "    \"permissionsRequiredTenant\" : [ \"okapi.proxy.self.timers.patch\" ], "
```
hjiebsco, line 350: `permissionsRequiredTenant` is used only here, no elsewhere. What does it do exactly?

okapi-core/src/main/raml/TimerDescriptor.json
```diff
@@ -0,0 +1,20 @@
...
+  "title": "TimerDescriptor",
+  "description": "Timer",
+  "type": "object",
+  "additionalProperties" : false,
+  "properties": {
+    "id": {
+      "description": "Timer ID",
+      "type": "string"
+    },
+    "routingEntry": {
```
hjiebsco, line 13: Not sure  if the name `routingEntry` makes sense in a `TimeDescriptor`.

doc/guide.md
adamdickmeiss: But the body would  not be a TimerDescriptor.json, then.. That would be a RoutingEntry.json.. Is that what you think is better?

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
adamdickmeiss: Yes. The changed timers are stored.

I'll make a call to reset a timer (to original value).

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
adamdickmeiss: I thought it wasn't necessaryu to use it elsewhere as this is same tenant and other operations are read-only.

If you ask what permissionsRequiredTenant is, then it is the effective permission used if the operating tenant is the same as the tenant in the path.. Eg if user can only modify timer in his/her library (self).. The other permission is used for a tenant that operates on ALL tenants.. Eg like a sysadmin..

okapi-core/src/main/raml/TimerDescriptor.json
adamdickmeiss: What should it be called, then?  "entry" , "handler" ?

doc/guide.md
adamdickmeiss: I can make this call as a option. It's a different path.. It's not really in harmondy with the GET which retrieves a TimerDescriptor.json..  So I'll keep the existing call and make this one - as another way.

okapi-core/src/main/java/org/folio/okapi/MainVerticle.java
adamdickmeiss: This is now done.

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
```diff
@@ -0,0 +1,312 @@
...
+
+public class TimerManager {
+
+  private final Logger logger = OkapiLogger.get();
+  private static final String TIMER_ENTRY_SEP = "_";
+  private static final String MAP_NAME = "org.folio.okapi.timer.map";
+  private static final String EVENT_NAME = "org.folio.okapi.timer.event";
+  private final LockedTypedMap2<TimerDescriptor> tenantTimers
+      = new LockedTypedMap2<>(TimerDescriptor.class);
+  private final Set<String> timerRunning = new HashSet<>();
```
hjiebsco, line 31: Does this need to be tenant id specific?

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
```diff
@@ -0,0 +1,312 @@
...
+          // if value has changed, then stop ..
+          if (!isSimilar(timerDescriptor, currentDescriptor)) {
+            return Future.succeededFuture();
+          }
+          // this timer is latest and current .. do the work..
+          // find module for this timer.. If module is not found, it was disabled
+          // in the meantime and timer is stopped.
+          return getModuleForTimer(tenantId, timerId).compose(md -> {
+            if (md == null) {
+              timerRunning.remove(timerId);
```
hjiebsco, line 182: If one tenant disable a timer modules, what does this `remove` mean for other tenants that still have the same time module enabled?

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
```diff
@@ -0,0 +1,312 @@
...
+          if (patchEntry.getDelay() == null && patchEntry.getUnit() == null
+              && patchEntry.getSchedule() == null) {
+            return tenantTimers.remove(tenantId, timerId).compose(deleted -> {
+              JsonObject o = new JsonObject();
+              o.put("tenantId", tenantId);
+              EventBus eb = vertx.eventBus();
+              eb.publish(EVENT_NAME, o.encode());
+              return Future.succeededFuture();
+            });
+          }
```
hjiebsco, line 266: Does this mean resetting one timer to default will trigger rescan all timers for a tenant?

hjiebsco (approved):
Left some minor comments/questions. BTW, looks like timer changes are not saved to db, so users have to be careful to restart Okapi clusters, correct? Maybe mention that in the guide.

hjiebsco:
BTW, I accidentally resolved the conversation about if `okapi.proxy.self.timers.patch` need to be defined as a regular perm or not. Please define it if needed.

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
adamdickmeiss: Yes. Will fix that

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
adamdickmeiss: Must be tenant specific

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
adamdickmeiss: I have changed the strategy and simplified, so that reset and patch is same path.

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
```diff
@@ -330,6 +337,30 @@ public static ModuleDescriptor moduleDescriptor(String okapiVersion) {
...
+        + "    \"methods\" :  [ \"GET\" ],"
+        + "    \"pathPattern\" : \"/_/proxy/tenants/{tenantId}/timers/{timerId}\","
+        + "    \"permissionsRequired\" : [ \"okapi.proxy.tenants.timers.get\" ], "
+        + "    \"permissionsRequiredTenant\" : [ ], "
+        + "    \"type\" : \"internal\" "
+        + "   }, {"
+        + "    \"methods\" :  [ \"PATCH\" ],"
+        + "    \"pathPattern\" : \"/_/proxy/tenants/{tenantId}/timers/{timerId}\","
+        + "    \"permissionsRequired\" : [  \"okapi.proxy.tenants.timers.patch\" ], "
+        + "    \"permissionsRequiredTenant\" : [ \"okapi.proxy.self.timers.patch\" ], "
```
hjiebsco, line 362: Is `okapi.proxy.self.timers.patch` defined?

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
```diff
@@ -0,0 +1,344 @@
...
+import org.folio.okapi.util.OkapiError;
+
+public class TimerManager {
+
+  private final Logger logger = OkapiLogger.get();
+  private static final String TIMER_ENTRY_SEP = "_";
+  private static final String MAP_NAME = "org.folio.okapi.timer.map";
+  private static final String EVENT_NAME = "org.folio.okapi.timer.event";
+  private final Map<String,LockedTypedMap1<TimerDescriptor>> tenantTimers = new HashMap<>();
+  private final Map<String,Long> timerRunning = new HashMap<>();
```
hjiebsco, line 34: Please update README to mention there is no db persistence for timer change.

adamdickmeiss: I'm about to add persistense too,

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
adamdickmeiss: Good.  Will add.

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
adamdickmeiss: Done.

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
```diff
@@ -0,0 +1,312 @@
...
+          }
+          // announce to shared map, then publish so that all instances of Okapi
+          // will get a new timer rolling
+          // the existing timer will notice that its timerDescriptor's routing entry is
+          // obsolete and terminate
+          return tenantTimers.put(tenantId, timerDescriptor.getId(), timerDescriptor)
+              .onSuccess(x -> {
+                JsonObject o = new JsonObject();
+                o.put("tenantId", tenantId);
+                o.put("timerDescriptor", Json.encode(timerDescriptor));
```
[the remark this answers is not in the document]
adamdickmeiss: > `Json.encode(timerDescriptor)` can be changed to `newJson` so not encode twice.

I have some brain damage. Can you be more specific?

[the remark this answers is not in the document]
hjiebsco: > > `Json.encode(timerDescriptor)` can be changed to `newJson` so not encode twice.
> 
> I have some brain damage. Can you be more specific?

I mean the encode part in these two lines look to be same.
Line 333 String newJson = Json.encode(newDescriptor);
Line 346 o.put("timerDescriptor", Json.encode(newDescriptor));

hjiebsco (approved):
LGTM. When disabling a module for a tenant, is there a parameter to remove timer changes for this tenant/module?

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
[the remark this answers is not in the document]
adamdickmeiss: Got it. Thx

adamdickmeiss:
> LGTM. When disabling a module for a tenant, is there a parameter to remove timer changes for this tenant/module?

No. You'd have to patch reset it.

PR #1112, opened 2021-04-09
Remove timers for disabled modules OKAPI-1005

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
```diff
@@ -78,6 +78,89 @@ private void tenantChange(String tenantId) {
...
 
+  private Future<Void> loadFromStorage(String tenantId) {
+    final LockedTypedMap1<TimerDescriptor> timerMap = tenantTimers.get(tenantId);
+    return timerStore.getAll().compose(list -> {
+          List<Future<Void>> futures = new LinkedList<>();
+          int prefixLen = tenantId.length() + 1;
+          for (TimerDescriptor timerDescriptor : list) {
+            String tenantTimerId = timerDescriptor.getId();
+            if (tenantTimerId.startsWith(tenantId + "_")) {
+              timerDescriptor.setId(tenantTimerId.substring(prefixLen));
```
julianladisch, line 89:
```suggestion
          String prefix = tenantId + TIMER_ENTRY_SEP;
           for (TimerDescriptor timerDescriptor : list) {
             String tenantTimerId = timerDescriptor.getId();
            if (tenantTimerId.startsWith(prefix)) {
              timerDescriptor.setId(tenantTimerId.substring(prefix.length()));
```

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
```diff
@@ -78,6 +78,89 @@ private void tenantChange(String tenantId) {
...
+      for (String timerId : list.keySet()) {
+        futures.add(getModuleForTimer(mdList, timerId).compose(md -> {
+          if (md != null) {
+            return Future.succeededFuture();
+          }
+          final String runId = tenantId + TIMER_ENTRY_SEP + timerId;
+          if (timerRunning.containsKey(runId)) {
+            vertx.cancelTimer(timerRunning.get(runId));
+            timerRunning.remove(runId);
+          }
```
julianladisch, line 113:
```suggestion
          Long id = timerRunning.remove(runId);
          if (id != null) {
            vertx.cancelTimer(id);
           }
```
Only a single call to timerRunning is needed.

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
```diff
@@ -178,23 +209,33 @@ private void fireTimer(String tenantId, ModuleDescriptor md, TimerDescriptor tim
...
 
   Future<ModuleDescriptor> getModuleForTimer(String tenantId, String timerId) {
-    return tenantManager.getEnabledModules(tenantId).compose(list -> {
-      String product = timerId.substring(0, timerId.indexOf(TIMER_ENTRY_SEP));
-      for (ModuleDescriptor md : list) {
-        if (product.equals(md.getProduct())) {
-          return Future.succeededFuture(md);
+    return tenantManager.getEnabledModules(tenantId)
+        .compose(list -> getModuleForTimer(list, timerId)
+        ).recover(cause -> Future.succeededFuture(null));
```
julianladisch, line 214:
```suggestion
        .compose(list -> getModuleForTimer(list, timerId))
        .recover(cause -> Future.succeededFuture(null));
```

okapi-core/src/main/java/org/folio/okapi/managers/TimerManager.java
```diff
@@ -178,23 +209,33 @@ private void fireTimer(String tenantId, ModuleDescriptor md, TimerDescriptor tim
...
+  }
+
+  Future<ModuleDescriptor> getModuleForTimer(List<ModuleDescriptor> list, String timerId) {
```
julianladisch, line 217: This should return a `ModuleDescriptor`, not a `Future<ModuleDescriptor>`, because there isn't any async.

Then
`.compose(list -> getModuleForTimer(list, timerId))` can be replaced by
`.map(list -> getModuleForTimer(list, timerId))`

julianladisch (approved):
Looks good, thanks!

PR #1115, opened 2021-04-12
OKAPI-876 Simplify CORS delegate and support preflight too

hjiebsco:
Please see https://issues.folio.org/browse/OKAPI-876. 
- Check CORS delegate just once to be efficient
- Allow preflight/OPTIONS to be delegated too

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -36,9 +36,9 @@ import org.folio.okapi.bean.ModuleDescriptor;
 import org.folio.okapi.bean.ModuleInstance;
 import org.folio.okapi.bean.RoutingEntry;
 import org.folio.okapi.bean.RoutingEntry.ProxyType;
-import org.folio.okapi.bean.Tenant;
 import org.folio.okapi.common.Config;
 import org.folio.okapi.common.ErrorType;
+import org.folio.okapi.common.HttpResponse;
 import org.folio.okapi.common.Messages;
 import org.folio.okapi.common.OkapiClient;
 import org.folio.okapi.common.OkapiLogger;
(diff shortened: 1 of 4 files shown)


adamdickmeiss (approved):
LGTM

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -1313,13 +1304,37 @@ public void redirectProxy(RoutingContext ctx) {
...
+        clientRequest.end();
+        clientRequest.response()
+            .onFailure(cause -> HttpResponse.responseError(ctx, 500, cause.getMessage()))
+            .onSuccess(res -> {
+              ctx.response().setStatusCode(res.statusCode());
+              ctx.response().headers().addAll(res.headers());
+              sanitizeAuthHeaders(ctx.response().headers());
+              ctx.response().end();
+            });
+      });
```
julianladisch, line 1328:
```suggestion
      httpClient.request(requestOptions)
          .compose(clientRequest -> {
            copyHeaders(clientRequest, ctx, null);
            clientRequest.end();
            return clientRequest.response();
          })
          .onSuccess(res -> {
            ctx.response().setStatusCode(res.statusCode());
            ctx.response().headers().addAll(res.headers());
            sanitizeAuthHeaders(ctx.response().headers());
            ctx.response().end();
          })
          .onFailure(cause -> HttpResponse.responseError(ctx, 500, cause.getMessage()));
```
Replacing onSuccess by compose avoids nesting and is easier to read and understand.
This also allows to drop the duplicate onFailure code.

okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
```diff
@@ -29,57 +30,60 @@ private CorsHelper() {
...
-  public static void addCorsHandler(Router router) {
+  public static void addCorsHandler(Router router, final TenantManager tenantManager) {
 
+    // set delegate CORS for special cases
     router.routeWithRegex("^/_/invoke/tenant/[^/ ]+/.*").handler(ctx -> {
-      ctx.data().put(CHECK_DELEGATE_CORS, true);
-      ctx.next();
+      String origPath = ctx.request().path();
+      String tenantId = origPath.replaceFirst("^/_/invoke/tenant/([^/ ]+)/.*$", "$1");
+      String newPath = origPath.replaceFirst("^/_/invoke/tenant/[^/ ]+(/.*$)", "$1");
```
julianladisch, line 39:
```
    router.routeWithRegex("^/_/invoke/tenant/([^/ ]+)(/.*)").handler(ctx -> {
      String tenantId = ctx.pathParam("param0");
      String newPath = ctx.pathParam("param1");
```
https://vertx.io/docs/vertx-web/java/#_capturing_path_parameters_with_regular_expressions

julianladisch (approved):
Looks good, thanks!

okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
hjiebsco: Nice! Thank you.

PR #1116, opened 2021-04-13
OKAPI-917: GET /_/version without permissions

julianladisch:


hjiebsco (commented):
Looks like this PR took a different approach from what described in https://issues.folio.org/browse/OKAPI-917. Just want to be sure every one is on the same page about removing the permission.

adamdickmeiss (approved):
I never understood why we had to protect this

julianladisch:
We use the solution #1119  instead.

PR #1117, opened 2021-04-19
Remove Future from return values

julianladisch:
If a method always returns a succeeded Future containing the result
the Future is no needed and should be removed.

Unneccessary Future wrapping bloates the code.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
@@ -227,35 +227,34 @@ public class TenantManager implements Liveness {
   Future<Void> enableAndDisableCheck(Tenant tenant, ModuleDescriptor modFrom,
                                      ModuleDescriptor modTo) {
 
-    return getEnabledModules(tenant).compose(modlist -> {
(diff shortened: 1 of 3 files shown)


okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -227,35 +227,34 @@ public void setProxyService(ProxyService px) {
...
+    HashMap<String, ModuleDescriptor> mods = new HashMap<>(modlist.size());
+    for (ModuleDescriptor md : modlist) {
+      mods.put(md.getId(), md);
+    }
+    if (modTo == null) {
       String deps = DepResolution.checkAllDependencies(mods);
-      if (!conflicts.isEmpty() || !deps.isEmpty()) {
-        return Future.failedFuture(new OkapiError(ErrorType.USER, conflicts + " " + deps));
+      if (!deps.isEmpty()) {
+        return Future.succeededFuture(); // failures even before we remove a module
```
julianladisch, line 238: Note that GitHub compares different lines! GitHub's comparison is misleading!

adamdickmeiss (approved):
thx

hjiebsco (approved):
Nice simplification!

PR #1119, opened 2021-04-23
OKAPI-917: GET /_/version with valid token, no permission needed

julianladisch:
Approach:
"permissionsRequiredTenant": []

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
+++ b/okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
@@ -388,7 +388,10 @@ public class RoutingEntry {
 
   /**
    * Match uri against routing entry with tenant.
-   * Method is NOT considered.
+   *
+   * <p>It also matches if pathPattern matches uri but pathPattern doesn't contain {tenantId}.
+   *
+   * <p>Method is NOT considered.
    *
    * @param uri actual path
    * @param tenant current tenant
--- a/okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
+++ b/okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
@@ -389,12 +389,13 @@ public class RoutingEntry {
   /**
    * Match uri against routing entry with tenant.
    * Method is NOT considered.
-   * @param uri acutal path
+   *
+   * @param uri actual path
    * @param tenant current tenant
-   * @return true if tenant matches tenantId in pathPattern; false otherwise
+   * @return true if uri matches pathPattern after replacing {tenantId} by tenant; false otherwise
    */
   public boolean matchUriTenant(String uri, String tenant) {
-    if (pathPattern == null || !pathPattern.contains("{tenantId}")) {
+    if (pathPattern == null) {
       return false;
     }
     String replacedPattern = pathPattern.replace("{tenantId}", tenant);
--- a/okapi-core/src/main/raml/RoutingEntry.json
+++ b/okapi-core/src/main/raml/RoutingEntry.json
@@ -81,7 +81,7 @@
       }
     },
     "permissionsRequiredTenant": {
-      "description": "Required permissions when tenant matches pathPattern",
+      "description": "Required permissions that replace permissionsRequired when the user is authenticated and either the tenant matches each {tenantId} in pathPattern or pathPattern doesn't contain {tenantId}",
       "type": "array",
       "items": {
         "type": "string"
--- a/okapi-core/src/test/java/org/folio/okapi/MultiTenantTest.java
+++ b/okapi-core/src/test/java/org/folio/okapi/MultiTenantTest.java
@@ -357,6 +357,20 @@ public class MultiTenantTest {
         .get("/_/proxy/tenants/" + tenant1 + "/modules")
         .then().statusCode(403).log().ifValidationFails();
 
+    // GET /_/version without token fails
+    given()
+        .header("Content-Type", "application/json")
+        .header("X-Okapi-Tenant", tenant2)
+        .get("/_/version")
+        .then().statusCode(401).log().ifValidationFails();
+
+    // GET /_/version with token succeeds: "permissionsRequiredTenant": []
+    given()
+        .header("Content-Type", "application/json")
+        .header("X-Okapi-Token", okapiTokenTenant2)
+        .get("/_/version")
+        .then().statusCode(200).log().ifValidationFails();
+
     // undeploy sample-module-1.2.0
     given()
       .header("Content-Type", "application/json")
--- a/okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
@@ -401,6 +401,7 @@ public class InternalModule {
         + "    \"methods\" :  [ \"GET\" ],"
         + "    \"pathPattern\" : \"/_/version\","
         + "    \"permissionsRequired\" : [ \"okapi.version.get\" ], "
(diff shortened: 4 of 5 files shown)


okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
```diff
@@ -389,12 +389,13 @@ static boolean fastMatch(String pathPattern, int patternI, String uri, int uriI,
...
-   * @param uri acutal path
+   *
+   * @param uri actual path
    * @param tenant current tenant
-   * @return true if tenant matches tenantId in pathPattern; false otherwise
+   * @return true if uri matches pathPattern after replacing {tenantId} by tenant; false otherwise
    */
   public boolean matchUriTenant(String uri, String tenant) {
-    if (pathPattern == null || !pathPattern.contains("{tenantId}")) {
+    if (pathPattern == null) {
```
adamdickmeiss, line 398: Ooops.. Now a tenant an access another tenants data..  permissionsRequiredTenant is now useless.

adamdickmeiss (changes requested):

okapi-core/src/test/java/org/folio/okapi/MultiTenantTest.java
```diff
@@ -357,6 +357,20 @@ public void test1(TestContext context) {
         .get("/_/proxy/tenants/" + tenant1 + "/modules")
         .then().statusCode(403).log().ifValidationFails();
 
+    // GET /_/version without token fails
+    given()
+        .header("Content-Type", "application/json")
+        .header("X-Okapi-Tenant", tenant2)
+        .get("/_/version")
+        .then().statusCode(401).log().ifValidationFails();
+
```
adamdickmeiss, line 366: Would also fail, had you used permissionsRequired instead of permissionsRequiredTenant.

okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
julianladisch: I don't follow.
This test
https://github.com/folio-org/okapi/pull/1119/files#diff-9b21a95ae0f6042b47f8f83988452dc68a22b3802d474e934d9719897f94142fL354-L358
shows that a tenant cannot access another tenant's data.
Can you add another unit tests that shows the issue you see?

okapi-core/src/test/java/org/folio/okapi/MultiTenantTest.java
julianladisch: Yes.
This test ensured that a token is required. This is different from the `/_/proxy/health` API that doesn't require a token.

okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
adamdickmeiss: If we relax on this, then one tenant can access another tenants data if there's no tenantId in the pattern. Perhaps that's fine, ie any good token will be able to get `/_/version`, since presumably that's not tenant specific anyway.

The description of permissionsRequiredTenant should be updated, then. Ie routingEntry.json description should be updated.. Not to mention, that the name is misleading.. 

Ie it should be permissionsRequiredWithToken.

okapi-core/src/test/java/org/folio/okapi/MultiTenantTest.java
adamdickmeiss: Right. I forgot that permissionsRequiredTenant is ONLY used when there's a good token involved.

adamdickmeiss (changes requested):
I think we should update the description in the schema. Otherwise, good.

okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
hjiebsco: Should we relax it, or maybe add a new endpoint to expose tenant specific Okapi version?

hjiebsco (commented):
Since UI can access API `/_/proxy/tenants/{tenantId}/modules`, it can get Okapi version from the response of that API call. It that works, there is no need to make changes to Okapi.

hjiebsco:
> Since UI can access API `/_/proxy/tenants/{tenantId}/modules`, it can get Okapi version from the response of that API call. It that works, there is no need to make changes to Okapi. @zburke, can you confirm?

zburke:
Hmmmm, it's an interesting idea, @hjiebsco. The response from that request looks like 
```
[
    ...
    {
        "id": "okapi-4.7.2"
    }
]
```
So we'd have to parse all the values in order to find the one that matches `/^okapi-{\d+\.}{3}$` or something similar. I'm not crazy about that, but it would be fine.

adamdickmeiss:
> Hmmmm, it's an interesting idea, @hjiebsco. The response from that request looks like
> 
> ```
> [
>     ...
>     {
>         "id": "okapi-4.7.2"
>     }
> ]
> ```
> 
> So we'd have to parse all the values in order to find the one that matches `/^okapi-{\d+\.}{3}$` or something similar. I'm not crazy about that, but it would be fine.

Be careful with regular expressions.. There could be a snapshot in there.. Take everything followed by - (if that follows by a digit).

hjiebsco:
> > Hmmmm, it's an interesting idea, @hjiebsco. The response from that request looks like
> > ```
> > [
> >     ...
> >     {
> >         "id": "okapi-4.7.2"
> >     }
> > ]
> > ```
> > 
> > 
> > So we'd have to parse all the values in order to find the one that matches `/^okapi-{\d+\.}{3}$` or something similar. I'm not crazy about that, but it would be fine.
> 
> Be careful with regular expressions.. There could be a snapshot in there.. Take everything followed by - (if that follows by a digit).
> 
> And for the record.. There's no point in protecting /_/version.. It was a mistake to do it in the first place.

Do we need to use regex? For value like `okapi-4.7.2`, wouldn't  stripping `okapi-` be enough? If it is a snapshot version like `okpai-4.8.0-SNAPSHOT`, we still want to keep the snapshot part, so it should be `4.8.0-SNAPSHOT`, no?

zburke:
So, in short, yes, we can get the okapi version from `/_/proxy/tenants/{tenantId}/modules` if we are willing to do some string manipulation. I don't like that the module name and version are concatenated but I'll get over it.

adamdickmeiss (approved):

hjiebsco (commented):
Just for the record. If we decide to use `permissionsRequiredTenant` to protect `/_/version`, I would prefer a new API `/_/proxy/tenants/{currentTenant}/version` as proposed in https://issues.folio.org/browse/OKAPI-917. This would be much more consistent. Also technically speaking UI reports modules enabled for a **tenant**.

okapi-core/src/main/java/org/folio/okapi/bean/RoutingEntry.java
julianladisch: routingEntry.json has been updated for the new use

julianladisch:
Merging as discussed and decided at the core platform meeting.

okapi-core/src/main/raml/RoutingEntry.json
```diff
@@ -81,7 +81,7 @@
       }
     },
     "permissionsRequiredTenant": {
-      "description": "Required permissions when tenant matches pathPattern",
+      "description": "Required permissions that replace permissionsRequired when the user is authenticated and either the tenant matches each {tenantId} in pathPattern or pathPattern doesn't contain {tenantId}",
```
hjiebsco, line 84: Just for the record, this field `permissionsRequiredTenant` and the new description look confusing (to me). It is also debatable whether this is a **breaking** change or not. Luckily only Okapi uses this at the moment.

PR #1124, opened 2021-05-20
OKAPI-1015 Fix saml login call timeout

hjiebsco:
Please see https://issues.folio.org/browse/OKAPI-1015. Basically Okapi 4.8.0 will time out when sending /saml/login request. I spent quite bit time debugging the issue, and found the workaround as shown in this PR. I am not sure why the old way breaks though. It would be great if someone can take a look and explain it. Thanks.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
@@ -6,7 +6,7 @@ import io.vertx.ext.web.Router;
 import io.vertx.ext.web.handler.CorsHandler;
 import java.util.List;
 import org.folio.okapi.bean.ModuleInstance;
-import org.folio.okapi.common.HttpResponse;
+import org.folio.okapi.bean.Tenant;
 import org.folio.okapi.common.XOkapiHeaders;
 import org.folio.okapi.managers.TenantManager;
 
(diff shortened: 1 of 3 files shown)


okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
```diff
@@ -39,20 +39,17 @@ public static void addCorsHandler(Router router, final TenantManager tenantManag
...
+      List<ModuleInstance> list = tenantManager.getModuleCache(t).lookup(newPath, method, moduleId);
+      for (ModuleInstance mi : list) {
+        if (mi.isHandler() && mi.getRoutingEntry().isDelegateCors()) {
+          ctx.data().put(DELEGATE_CORS, true);
+          ctx.data().put(DELEGATE_CORS_MODULE_INSTANCE, mi);
+          break;
+        }
+      }
+      ctx.next();
     });
```
adamdickmeiss, line 53: The original code returns immediately .. and ctx is changed in the background. It's not the full handler and does not "call" anything afterwards... 

```
 //handle CORS
    CorsHelper.addCorsHandler(router, tenantManager);

    if (proxyService != null) {
      router.routeWithRegex("^/_/invoke/tenant/[^/ ]+/.*")
          .handler(proxyService::redirectProxy);
      // Note: This can not go into the InternalModule, it reads the req body,
      // and then we can not ctx.reroute(). Unless we do something trickier,
      // like a new HTTP request.
    }

``` 

The /_/invoke/tenant should be handled in ONE handler.. so this code should in the redirectProxy code (or sub thereof). I also don't like the DELEGATE_CORS_MODULE_INSTANCE.. That's not needed when handled in one place. -- finally - this code (like redirectProxy) should only be in use when proxyService != null

okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
```diff
@@ -39,20 +39,17 @@ public static void addCorsHandler(Router router, final TenantManager tenantManag
...
-                ctx.data().put(DELEGATE_CORS, true);
-                ctx.data().put(DELEGATE_CORS_MODULE_INSTANCE, mi);
-                break;
-              }
-            }
-            ctx.next();
-          });
+      Tenant t = new Tenant();
+      t.getDescriptor().setId(tenantId);
+      List<ModuleInstance> list = tenantManager.getModuleCache(t).lookup(newPath, method, moduleId);
```
adamdickmeiss, line 44: Let's leave things as they are (mostly) for now.. Can you make getModuleCache take a string rather than a Tenant class? That will make things a little prettier. Thx

adamdickmeiss (changes requested):

okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
hjiebsco: > The original code returns immediately .. and ctx is changed in the background. It's not the full handler and does not "call" anything afterwards...

That makes sense. Thanks. The error message (Connection was closed to mod-login-saml) was kind misleading. While debugging it, I can see calls the right mod-login-saml url with the right headers and content. I was puzzled. :)

hjiebsco: >The /_/invoke/tenant should be handled in ONE handler.. so this code should in the redirectProxy code (or sub thereof). I also don't like the DELEGATE_CORS_MODULE_INSTANCE.. That's not needed when handled in one place. -- finally - this code (like redirectProxy) should only be in use when proxyService != null

This will need bigger changes due to sequence of those handlers. If I understand your other message correct (`Let's leave things as they are (mostly) for now.. `), we will not change it in this PR. Let me know otherwise.

okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
hjiebsco: > Let's leave things as they are (mostly) for now.. Can you make getModuleCache take a string rather than a Tenant class? That will make things a little prettier. Thx

Done.

adamdickmeiss (approved):

PR #1134, opened 2021-06-21
OKAPI-1023: Pull module descriptors with bulk, fix connection timeout

julianladisch:
### Overview:

Connection timeout when PullManager writes > 14000 module descriptors
from folio-registry.dev.folio.org into PostgreSQL.

### Steps to Reproduce:

* Run Okapi with empty database
* curl -w '\n' -D - -X POST -H "Content-type: application/json" -d '{ "urls": ["https://folio-registry.dev.folio.org"] }' http://localhost:9130/_/proxy/pull/modules

### Expected Results:

Successfully inserts > 14000 module descriptors into PostgreSQL

### Actual Results:

```
17:43:01 [239139/proxy] [supertenant] [] [] INFO  PullManager          pull: 14313 MDs to insert
17:43:35 [239139/proxy] [supertenant] [] [] WARN  ProxyContext         Timeout
```

And some module descriptors are NOT written to the database.

### Additional Information:

This is the current implementation:
https://github.com/folio-org/okapi/blob/v4.8.2/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java#L122-L131
https://github.com/folio-org/okapi/blob/v4.8.2/okapi-core/src/main/java/org/folio/okapi/service/impl/PostgresTable.java#L52-L61
https://github.com/folio-org/okapi/blob/v4.8.2/okapi-core/src/main/java/org/folio/okapi/service/impl/PostgresQuery.java#L31
For 14313 module descriptors 14313 Futures are started in parallel. Each Future waits to get a connection, writes one module descriptor, and closes the connection.
The default connection timeout is 30 seconds:
https://github.com/eclipse-vertx/vertx-sql-client/blob/4.1.0/vertx-sql-client/src/main/java/io/vertx/sqlclient/PoolOptions.java#L61
On some servers executing all Futures takes more than 30 seconds and the futures that waited more than 30 seconds fail with a timeout error.

### Solution:

Use a single database connection with a single batch/bulk command to insert
all module descriptors in one go. This will never timeout because a
running query doesn't timeout, only waiting for a free connection times out.

This solution speeds up the PostgreSQL insert from 40 seconds to 12 seconds.

### DIFF
--- a/okapi-common/pom.xml
+++ b/okapi-common/pom.xml
@@ -56,9 +56,8 @@
       <version>1.13</version>
     </dependency>
     <dependency>
-      <groupId>com.jayway.restassured</groupId>
+      <groupId>io.rest-assured</groupId>
       <artifactId>rest-assured</artifactId>
-      <version>2.8.0</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/okapi-core/pom.xml
+++ b/okapi-core/pom.xml
@@ -80,7 +80,6 @@
     <dependency>
       <groupId>io.rest-assured</groupId>
       <artifactId>rest-assured</artifactId>
-      <version>3.0.7</version>
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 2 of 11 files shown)


adamdickmeiss (approved):
thx

PR #1135, opened 2021-06-22
Facility to remove obsolete modules OKAPI-1024

adamdickmeiss:
New service `/_/proxy/cleanup/modules?saveReleases=`N`&saveSnapshots=`M

where
 N is the number of latest releases where snapshots are never considered obsolete
 M is the number of latest snapshots that are never considered obsolete (preserved)

Remove method Future<Void> insert(ModuleDescriptor md) from
ModuleStore interface as it's no longer in use with OKAPI-1023
work.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
@@ -20,6 +20,7 @@ import org.folio.okapi.common.OkapiLogger;
 import org.folio.okapi.service.ModuleStore;
 import org.folio.okapi.util.DepResolution;
 import org.folio.okapi.util.LockedTypedMap1;
+import org.folio.okapi.util.ModuleUtil;
 import org.folio.okapi.util.OkapiError;
 
 /**
@@ -57,9 +58,6 @@ public class ModuleManager {
    * @return future result
    */
   private Future<Void> loadModules() {
-    if (moduleStore == null) {
-      return Future.succeededFuture();
-    }
     return modules.size().compose(kres -> {
       if (kres > 0) {
         logger.debug("Not loading modules, looks like someone already did");
@@ -83,10 +81,12 @@ public class ModuleManager {
    * @param preRelease whether to allow pre-releasee
    * @param npmSnapshot whether to allow npm-snapshot
    * @param removeIfMissingDep skip modules where dependency check fails
+   * @param deleteObsolete if true: delete obsolete modules
    * @return future
    */
   public Future<Void> createList(List<ModuleDescriptor> list, boolean check, boolean preRelease,
-                                 boolean npmSnapshot, boolean removeIfMissingDep) {
+                                 boolean npmSnapshot, boolean removeIfMissingDep,
+                                 boolean deleteObsolete) {
     return getModulesWithFilter(preRelease, npmSnapshot, null).compose(ares -> {
       Map<String, ModuleDescriptor> tempList = new HashMap<>();
       for (ModuleDescriptor md : ares) {
@@ -115,15 +115,31 @@ public class ModuleManager {
           return Future.failedFuture(new OkapiError(ErrorType.USER, res));
         }
       }
-      return createList2(newList);
+      return createList2(newList).compose(x -> {
+        if (deleteObsolete) {
+          return deleteObsolete();
+        } else {
+          return Future.succeededFuture();
+        }
+      });
     });
   }
 
+  Future<Void> deleteObsolete() {
+    return modules.getAll()
+        .compose(ares -> {
+          List<ModuleDescriptor> newList = new LinkedList<>(ares.values());
+          Future<Void> future = Future.succeededFuture();
+          for (ModuleDescriptor md: ModuleUtil.getObsolete(newList)) {
+            future = future.compose(x -> moduleStore.delete(md.getId()).mapEmpty());
+            future = future.compose(x -> modules.remove(md.getId()).mapEmpty());
+          }
+          return future;
+        });
+  }
+
   private Future<Void> createList2(List<ModuleDescriptor> list) {
-    Future<Void> storeFuture = Future.succeededFuture();
-    if (moduleStore != null) {
-      storeFuture = moduleStore.insert(list);
-    }
+    Future<Void> storeFuture = moduleStore.insert(list);
     List<Future<Void>> futures = new LinkedList<>();
     for (ModuleDescriptor md : list) {
       futures.add(modules.add(md.getId(), md));
@@ -142,13 +158,7 @@ public class ModuleManager {
   public Future<Void> delete(String id) {
     return modules.getAll()
         .compose(ares -> deleteCheckDep(id, ares))
-        .compose(res -> {
-          if (moduleStore == null) {
-            return Future.succeededFuture();
-          } else {
-            return moduleStore.delete(id).mapEmpty();
-          }
-        })
+        .compose(res -> moduleStore.delete(id).mapEmpty())
         .compose(res -> deleteInternal(id).mapEmpty());
   }
 
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
@@ -117,12 +117,29 @@ public class ModuleManager {
     });
   }
 
-  Future<Void> deleteObsolete(int saveReleases, int saveSnapshots) {
+  Future<Void> deleteObsolete(Map<String, String> inUse, int saveReleases, int saveSnapshots) {
     return modules.getAll()
         .compose(ares -> {
           List<ModuleDescriptor> newList = new LinkedList<>(ares.values());
+          List<ModuleDescriptor> obsolete =
+              ModuleUtil.getObsolete(newList, saveReleases, saveSnapshots);
+          for (ModuleDescriptor md: obsolete) {
+            if (inUse.containsKey(md.getId())) {
+              return Future.failedFuture(new OkapiError(ErrorType.USER,
+                  messages.getMessage("10206", md.getId(), inUse.get(md.getId()))));
+            }
+            ares.remove(md.getId());
+          }
+          for (ModuleDescriptor md: obsolete) {
+            ares.remove(md.getId());
+          }
+          String res = DepResolution.checkAllDependencies(ares);
+          if (!res.isEmpty()) {
+            return Future.failedFuture(new OkapiError(ErrorType.USER,
+                messages.getMessage("10212", res)));
+          }
           Future<Void> future = Future.succeededFuture();
-          for (ModuleDescriptor md: ModuleUtil.getObsolete(newList, saveReleases, saveSnapshots)) {
+          for (ModuleDescriptor md: obsolete) {
             future = future.compose(x -> moduleStore.delete(md.getId()).mapEmpty());
             future = future.compose(x -> modules.remove(md.getId()).mapEmpty());
           }
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
@@ -167,8 +167,9 @@ public class ModuleManager {
   public Future<Void> delete(String id) {
     return modules.getAll()
         .compose(ares -> deleteCheckDep(id, ares))
-        .compose(res -> moduleStore.delete(id).mapEmpty())
-        .compose(res -> deleteInternal(id).mapEmpty());
+        .compose(res -> moduleStore.delete(id))
+        .compose(res -> deleteInternal(id))
+        .mapEmpty();
   }
 
   private Future<Void> deleteCheckDep(String id, LinkedHashMap<String, ModuleDescriptor> mods) {
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
@@ -81,12 +81,10 @@ public class ModuleManager {
    * @param preRelease whether to allow pre-releasee
    * @param npmSnapshot whether to allow npm-snapshot
    * @param removeIfMissingDep skip modules where dependency check fails
-   * @param deleteObsolete if true: delete obsolete modules
    * @return future
    */
   public Future<Void> createList(List<ModuleDescriptor> list, boolean check, boolean preRelease,
-                                 boolean npmSnapshot, boolean removeIfMissingDep,
-                                 boolean deleteObsolete) {
+                                 boolean npmSnapshot, boolean removeIfMissingDep) {
     return getModulesWithFilter(preRelease, npmSnapshot, null).compose(ares -> {
       Map<String, ModuleDescriptor> tempList = new HashMap<>();
       for (ModuleDescriptor md : ares) {
@@ -115,27 +113,8 @@ public class ModuleManager {
           return Future.failedFuture(new OkapiError(ErrorType.USER, res));
         }
       }
-      return createList2(newList).compose(x -> {
-        if (deleteObsolete) {
-          return deleteObsolete();
-        } else {
-          return Future.succeededFuture();
-        }
-      });
+      return createList2(newList);
     });
-  }
-
-  Future<Void> deleteObsolete() {
-    return modules.getAll()
-        .compose(ares -> {
-          List<ModuleDescriptor> newList = new LinkedList<>(ares.values());
-          Future<Void> future = Future.succeededFuture();
-          for (ModuleDescriptor md: ModuleUtil.getObsolete(newList, 1, 0)) {
-            future = future.compose(x -> moduleStore.delete(md.getId()).mapEmpty());
-            future = future.compose(x -> modules.remove(md.getId()).mapEmpty());
-          }
-          return future;
-        });
   }
 
   Future<Void> deleteObsolete(int saveReleases, int saveSnapshots) {
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
@@ -140,8 +140,8 @@ public class ModuleManager {
           }
           Future<Void> future = Future.succeededFuture();
           for (ModuleDescriptor md: obsolete) {
-            future = future.compose(x -> moduleStore.delete(md.getId()).mapEmpty());
-            future = future.compose(x -> modules.remove(md.getId()).mapEmpty());
+            future = future.compose(x -> moduleStore.delete(md.getId()))
+                           .compose(x -> modules.remove(md.getId())).mapEmpty();
           }
           return future;
         });
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
@@ -130,7 +130,20 @@ public class ModuleManager {
         .compose(ares -> {
           List<ModuleDescriptor> newList = new LinkedList<>(ares.values());
           Future<Void> future = Future.succeededFuture();
-          for (ModuleDescriptor md: ModuleUtil.getObsolete(newList)) {
+          for (ModuleDescriptor md: ModuleUtil.getObsolete(newList, 1, 0)) {
+            future = future.compose(x -> moduleStore.delete(md.getId()).mapEmpty());
+            future = future.compose(x -> modules.remove(md.getId()).mapEmpty());
+          }
+          return future;
+        });
+  }
+
+  Future<Void> deleteObsolete(int saveReleases, int saveSnapshots) {
+    return modules.getAll()
+        .compose(ares -> {
+          List<ModuleDescriptor> newList = new LinkedList<>(ares.values());
+          Future<Void> future = Future.succeededFuture();
+          for (ModuleDescriptor md: ModuleUtil.getObsolete(newList, saveReleases, saveSnapshots)) {
             future = future.compose(x -> moduleStore.delete(md.getId()).mapEmpty());
             future = future.compose(x -> modules.remove(md.getId()).mapEmpty());
           }
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
@@ -3,6 +3,7 @@ package org.folio.okapi.managers;
 import io.vertx.core.Future;
 import io.vertx.core.Vertx;
 import io.vertx.core.json.Json;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
@@ -90,7 +91,7 @@ public class ModuleManager {
       for (ModuleDescriptor md : ares) {
         tempList.put(md.getId(), md);
       }
-      LinkedList<ModuleDescriptor> newList = new LinkedList<>();
+      List<ModuleDescriptor> newList = new LinkedList<>();
       for (ModuleDescriptor md : list) {
         final String id = md.getId();
         if (tempList.containsKey(id)) {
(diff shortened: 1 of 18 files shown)


hjiebsco (commented):
Looks fine to me. Two minor questions/suggestions:

- If enable this new `deleteObsolete` parameter for `/proxy/modules` POST call, that would mean one can post MD for modA and at same time clean up obsolete MDs for modB, modC, and etc. Is this a little bit counter intuitive? Would it be more straightforward to add a separate API to clean up obsolete MDs? You can use different parameters to control deletion of not just snapshot version, but also possibly unused stable versions (to keep Okapi lean for example).
- Looks to me the test added to ProxyTest does not really check anything about removing obsolete MDs. Maybe that is not necessary?

hjiebsco (approved):

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -115,15 +115,31 @@ public ModuleManager(ModuleStore moduleStore, boolean local) {
...
         }
       }
-      return createList2(newList);
+      return createList2(newList).compose(x -> {
+        if (deleteObsolete) {
+          return deleteObsolete();
+        } else {
+          return Future.succeededFuture();
+        }
+      });
```
julianladisch, line 124:
```suggestion
      Future<Void> future = createList2(newList);
      if (deleteObsolete) {
        future = future.compose(deleteObsolete());
      }
      return future;
```

okapi-core/src/main/java/org/folio/okapi/util/ModuleUtil.java
```diff
@@ -167,4 +168,37 @@ public static String moduleList(List<ModuleDescriptor> moduleDescriptors) {
...
+  /**
+   * Return list of obsolete modules.
+   *
+   * <p>A module is considered obsolete if it's a snapshot *and* there is a higher
+   * non-snapshot version.
+   * @param mdl list of modules to consider
+   * @return list of obsolete modules.
+   */
+  public static List<ModuleDescriptor> getObsolete(List<ModuleDescriptor> mdl) {
+    mdl.sort(Collections.reverseOrder());
```
julianladisch, line 181: This sorts the input list in descending order. This should be documented in the javadoc so that the caller knows to pass a modifiable list and can make a copy if the old order is still needed.

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -115,15 +115,31 @@ public ModuleManager(ModuleStore moduleStore, boolean local) {
...
     });
   }
 
+  Future<Void> deleteObsolete() {
+    return modules.getAll()
+        .compose(ares -> {
+          List<ModuleDescriptor> newList = new LinkedList<>(ares.values());
```
julianladisch, line 131: The LinkedList should be replaced by an ArrayList for faster sorting in getObsolete.

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -115,15 +115,31 @@ public ModuleManager(ModuleStore moduleStore, boolean local) {
...
+          Future<Void> future = Future.succeededFuture();
+          for (ModuleDescriptor md: ModuleUtil.getObsolete(newList)) {
+            future = future.compose(x -> moduleStore.delete(md.getId()).mapEmpty());
+            future = future.compose(x -> modules.remove(md.getId()).mapEmpty());
```
julianladisch, line 135:
```suggestion
            future = future.compose(x -> moduleStore.delete(md.getId()))
                         .compose(x -> modules.remove(md.getId()))
                         .mapEmpty();
```

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -115,15 +115,31 @@ public ModuleManager(ModuleStore moduleStore, boolean local) {
...
+      });
     });
   }
 
+  Future<Void> deleteObsolete() {
+    return modules.getAll()
+        .compose(ares -> {
+          List<ModuleDescriptor> newList = new LinkedList<>(ares.values());
+          Future<Void> future = Future.succeededFuture();
+          for (ModuleDescriptor md: ModuleUtil.getObsolete(newList)) {
```
julianladisch, line 133: How can we ensure that a md is not deleted when the module is currently deployed?

okapi-core/src/main/java/org/folio/okapi/util/ModuleUtil.java
```diff
@@ -167,4 +168,37 @@ public static String moduleList(List<ModuleDescriptor> moduleDescriptors) {
     }
     return str.toString();
   }
+
+  /**
+   * Return list of obsolete modules.
+   *
+   * <p>A module is considered obsolete if it's a snapshot *and* there is a higher
+   * non-snapshot version.
```
julianladisch, line 176: This assumes that we maintain only the branch of the highest version.
We might maintain a branch for Iris with bugfixes while maintaining a branch for Juniper.

My suggestion:
A module is considered obsolete if it's a snapshot and
a) a higher snapshot exists that differs by the snapshot version only (compareTo is -1 or 1), or
b) at least five higher non-snapshot versions exist.

okapi-core/src/main/raml/okapi.raml
```diff
@@ -354,6 +354,11 @@ types:
         type: boolean
         required: false
         default: true
+      deleteObsolete:
+        description: Whether to remove obsolete modules as well.
```
julianladisch, line 358: Should explain what "obsolete" means.

okapi-core/src/main/raml/okapi.raml
```diff
@@ -397,6 +402,11 @@ types:
         type: boolean
         required: false
         default: true
+      deleteObsolete:
+        description: Whether to remove obsolete modules as well
```
julianladisch, line 406: Should explain what "obsolete" means.

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,35 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
...
 
+  @Test
+  public void testObsoleteModules(TestContext context) {
+    RestAssuredClient c;
+
+    ModuleDescriptor mdA = new ModuleDescriptor();
+    mdA.setId("moduleA-1.0.0-SNAPSHOT.1");
+
+    List<ModuleDescriptor> modules = new LinkedList<>();
+    modules.add(mdA);
```
julianladisch, line 4548:
```suggestion
    List<ModuleDescriptor> modules = List.of(mdA);
```

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,35 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
...
+    c.given()
+        .header("Content-Type", "application/json")
+        .body(Json.encodePrettily(modules)).post("/_/proxy/import/modules?deleteObsolete=true")
+        .then().statusCode(204);
+    Assert.assertTrue("raml: " + c.getLastReport().toString(),
+        c.getLastReport().isEmpty());
+
+    mdA.setId("moduleA-1.0.0");
+    modules = new LinkedList<>();
+    modules.add(mdA);
```
julianladisch, line 4559:
```suggestion
    modules = List.of(mdA);
```

okapi-core/src/test/java/org/folio/okapi/util/ModuleUtilTest.java
```diff
@@ -154,4 +154,64 @@ void testModuleList() {
     list.add(md);
     assertThat(ModuleUtil.moduleList(list)).isEqualTo("foo-1.0.0, bar-2.0.0");
   }
+
+  @Test
+  void testGetObsolete1() {
+    assertThat(modulesList.size()).isEqualTo(104);
```
julianladisch, line 160:
```suggestion
    assertThat(modulesList).hasSize(104);
```

okapi-core/src/test/java/org/folio/okapi/util/ModuleUtilTest.java
```diff
@@ -154,4 +154,64 @@ void testModuleList() {
...
+    List<ModuleDescriptor> obsolete = ModuleUtil.getObsolete(modulesList);
+    assertThat(obsolete.isEmpty()).isTrue();
```
julianladisch, line 162:
```suggestion
    assertThat(obsolete).isEmpty();
```

okapi-core/src/test/java/org/folio/okapi/util/ModuleUtilTest.java
```diff
@@ -154,4 +154,64 @@ void testModuleList() {
...
+        "mod-b-1.0.0-SNAPSHOT.1",
+        "mod-a-1.1.0-SNAPSHOT.5",
+        "mod-a-0.8.0-SNAPSHOT.1",
+        "mod-a-0.8.0-SNAPSHOT.2",
+        "mod-a-0.8.0",
+    };
+    List<ModuleDescriptor> mds = new LinkedList<>();
+    for (int i = 0; i < ids.length; i++) {
+      ModuleDescriptor md = new ModuleDescriptor();
+      md.setId(ids[i]);
```
julianladisch, line 180:
```suggestion
    for (String id : ids) {
      ModuleDescriptor md = new ModuleDescriptor();
      md.setId(id);
```

okapi-core/src/test/java/org/folio/okapi/util/ModuleUtilTest.java
```diff
@@ -154,4 +154,64 @@ void testModuleList() {
...
+      mds.add(md);
+    }
+    assertThat(mds.size()).isEqualTo(8);
```
julianladisch, line 183: This assert is not needed. It doesn't test the class under test but our test setup.

okapi-core/src/test/java/org/folio/okapi/util/ModuleUtilTest.java
```diff
@@ -154,4 +154,64 @@ void testModuleList() {
...
+    List<ModuleDescriptor> obsolete = ModuleUtil.getObsolete(mds);
+    assertThat(obsolete.size()).isEqualTo(4);
```
julianladisch, line 185:
```suggestion
    assertThat(obsolete).hasSize(4);
```

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,35 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
         .then().statusCode(200).log().ifValidationFails();
   }
 
+  @Test
+  public void testObsoleteModules(TestContext context) {
```
julianladisch, line 4541: After the POST a GET /_/proxy/modules should be called to check that some obsolete module has been removed.

skoczko:
@wafschneider @funkymalc Guys, comments regarding this feature would be appreciated!

okapi-core/src/main/java/org/folio/okapi/util/ModuleUtil.java
adamdickmeiss: Yes. I was not aware of snapshots in use for bug fix releases or branches therein. That changes the picture , quite a lot. So IMHO we should just not bother with this feature and continue to have all modules in one pile.

julianladisch:
Keeping N snapshots will delete snapshots that are in use. If N=100 and there are 100 snapshots for 5.0.0-* the latest snapshot of 4.3.0-* will be deleted even if we still maintain and use 4.3.0-*.

My suggestion: A module is considered obsolete if it's a snapshot and
a) a higher snapshot exists that differs by the snapshot version only (compareTo is -1 or 1, for example 4.3.0-snapshot-123 allows to delete 4.3.0-snapshot-122), or
b) at least five higher non-snapshot versions exist (if there is 5.0.0, 4.3.0, 4.2.3, 4.2.2, 4.2.1 we can delete all snapshots for <= 4.2.0-*).

skoczko:
@wafschneider @funkymalc Guys, any comments here? I think we should either go with Adam's strategy (keep only snapshots for the latest release) or, and I would prefer that, with a simpler strategy that just keeps N latest snapshots (no matter the release).

funkymalc:
> Keeping N snapshots will delete snapshots that are in use. If N=100 and there are 100 snapshots for 5.0.0-* the latest snapshot of 4.3.0-* will be deleted even if we still maintain and use 4.3.0-*.
> 
> My suggestion: A module is considered obsolete if it's a snapshot and
> a) a higher snapshot exists that differs by the snapshot version only (compareTo is -1 or 1, for example 4.3.0-snapshot-123 allows to delete 4.3.0-snapshot-122), or
> b) at least five higher non-snapshot versions exist (if there is 5.0.0, 4.3.0, 4.2.3, 4.2.2, 4.2.1 we can delete all snapshots for <= 4.2.0-*).

I think this may be the best strategy.  Often times a major or minor version of a snapshot is changed and committed to master (snapshot),  but a particular build may use the previous minor or major snapshot version for interface dependency reasons.

wafschneider:
I think I agree with @funkymalc, as lovely as it would be to be very strict about it.

Regarding @hjiebsco's comment:
>If enable this new deleteObsolete parameter for /proxy/modules POST call, that would mean one can post MD for modA and at same time clean up obsolete MDs for modB, modC, and etc.

Can we clarify the behavior? I think it would be surprising to act on more than one module at a time with this query parameter -- i.e., if you are registering a new version of modA, I would _not_ expect to purge descriptors for modB.

Finally, I do have a worry about unexpectedly deleting snapshot module versions that are still use in reference environments from the registry (less concern about snapshot versions in use in scratch environments, as their Okapi instance is not rebuilt nightly) -- but I think @julianladisch's suggestion makes that unlikely.

skoczko:
We decided to make a seperate cleanup webservice that will take two parameters:

N -- number of releases to save
M -- number of snapshots to save for each release

okapi-core/src/main/raml/okapi.raml
```diff
@@ -386,6 +386,42 @@ types:
...
 
+/_/proxy/cleanup/modules:
+  description: Proxy modules service, clean up obsolete snapshot modules
+  post:
+    description: Remove module snapshot modules descriptors that are obsolete. A module is
+      considered obsolete if it is part of a release that is earlier than N latest releases (saveReleases) and
+      not part of the M latest snapshots in that release.
+    queryParameters:
+      saveReleases:
+        description: The latest releases where snapshots are never considered obsolete
```
julianladisch, line 397:
```suggestion
        description: The number of latest releases where snapshots are never considered obsolete
```

okapi-core/src/main/raml/okapi.raml
```diff
@@ -386,6 +386,42 @@ types:
...
+        type: integer
+        required: true
+      saveSnapshots:
+        description: The latest snapshots that are never considered obsolete (preserved)
```
julianladisch, line 401:
```suggestion
        description: The number of latest snapshots that are never considered obsolete (preserved)
```

okapi-core/src/main/raml/okapi.raml
```diff
@@ -386,6 +386,42 @@ types:
...
+        type: integer
+        required: true
+    body:
+      application/json:
```
julianladisch, line 405: The body is not used and should not be declared in the .raml.

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,83 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
         .then().statusCode(200).log().ifValidationFails();
   }
 
+  @Test
+  public void testObsoleteModules2(TestContext context) {
+    RestAssuredClient c;
+
+    given()
+        .header("Content-Type", "application/json")
+        .body("{}").post("/_/proxy/cleanup/modules?saveSnapshots=2")
```
julianladisch, line 4546:
```suggestion
        .post("/_/proxy/cleanup/modules?saveSnapshots=2")
```
The body is not used and is ignored.

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,83 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
...
+        .then().statusCode(400)
+        .body(is("Missing value for parameter 'saveReleases'"));
+    given()
+        .header("Content-Type", "application/json")
+        .body("{}").post("/_/proxy/cleanup/modules?saveReleases=1")
```
julianladisch, line 4551:
```suggestion
        .post("/_/proxy/cleanup/modules?saveReleases=1")
```
The body is not used and is ignored.

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,83 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
...
+        .then().statusCode(400)
+        .body(is("Missing value for parameter 'saveSnapshots'"));
+    c = api.createRestAssured3();
+    c.given()
+        .header("Content-Type", "application/json")
+        .body("{}").post("/_/proxy/cleanup/modules?saveReleases=1&saveSnapshots=2")
```
julianladisch, line 4557:
```suggestion
        .post("/_/proxy/cleanup/modules?saveReleases=1&saveSnapshots=2")
```
The body is not used and is ignored, even when passing in some invalid JSON like "}".

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,83 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
...
+        .body("$", hasSize(4))
+        .body("[0].id", is("moduleA-1.0.0-SNAPSHOT.1"))
+        .body("[1].id", is("moduleA-1.0.0-SNAPSHOT.2"));
+    Assert.assertTrue("raml: " + c.getLastReport().toString(),
+        c.getLastReport().isEmpty());
+
+    c = api.createRestAssured3();
+    c.given()
+        .header("Content-Type", "application/json")
+        .body("{}").post("/_/proxy/cleanup/modules?saveReleases=0&saveSnapshots=1")
```
julianladisch, line 4602:
```suggestion
        .post("/_/proxy/cleanup/modules?saveReleases=0&saveSnapshots=1")
```
The body is not used and is ignored.

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,83 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
...
+        .body("{}").post("/_/proxy/cleanup/modules?saveReleases=1")
+        .then().statusCode(400)
+        .body(is("Missing value for parameter 'saveSnapshots'"));
+    c = api.createRestAssured3();
+    c.given()
+        .header("Content-Type", "application/json")
+        .body("{}").post("/_/proxy/cleanup/modules?saveReleases=1&saveSnapshots=2")
+        .then().statusCode(204);
+    Assert.assertTrue("raml: " + c.getLastReport().toString(),
+        c.getLastReport().isEmpty());
```
julianladisch, line 4560:
```suggestion
    api.failFast().createRestAssured3()
    .given()
        .header("Content-Type", "application/json")
        .body("{}").post("/_/proxy/cleanup/modules?saveReleases=1&saveSnapshots=2")
        .then().statusCode(204);
```
The Assert can be replaced by failFast().

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,83 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
...
+    mdA.setId("moduleA-1.0.0");
+    modules.add(mdA);
+
+    c = api.createRestAssured3();
+    c.given()
+        .header("Content-Type", "application/json")
+        .body(Json.encodePrettily(modules)).post("/_/proxy/import/modules")
+        .then().statusCode(204);
+    Assert.assertTrue("raml: " + c.getLastReport().toString(),
+        c.getLastReport().isEmpty());
```
julianladisch, line 4587: Assert is not needed when using api.failFast().

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,83 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
...
+
+    c = api.createRestAssured3();
+    c.given()
+        .get("/_/proxy/modules")
+        .then().statusCode(200)
+        .body("$", hasSize(4))
+        .body("[0].id", is("moduleA-1.0.0-SNAPSHOT.1"))
+        .body("[1].id", is("moduleA-1.0.0-SNAPSHOT.2"));
+    Assert.assertTrue("raml: " + c.getLastReport().toString(),
+        c.getLastReport().isEmpty());
```
julianladisch, line 4597: Assert is not needed when using api.failFast().

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,83 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
...
+
+    c = api.createRestAssured3();
+    c.given()
+        .header("Content-Type", "application/json")
+        .body("{}").post("/_/proxy/cleanup/modules?saveReleases=0&saveSnapshots=1")
+        .then().statusCode(204);
+    Assert.assertTrue("raml: " + c.getLastReport().toString(),
+        c.getLastReport().isEmpty());
```
julianladisch, line 4605: Assert is not needed when using api.failFast().

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4537,6 +4537,83 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
...
+
+    c = api.createRestAssured3();
+    c.given()
+        .get("/_/proxy/modules")
+        .then().statusCode(200)
+        .body("$", hasSize(3))
+        .body("[0].id", is("moduleA-1.0.0-SNAPSHOT.2"));
+    Assert.assertTrue("raml: " + c.getLastReport().toString(),
+        c.getLastReport().isEmpty());
```
julianladisch, line 4614: Assert is not needed when using api.failFast().

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
adamdickmeiss: done now

okapi-core/src/main/java/org/folio/okapi/managers/ModuleManager.java
```diff
@@ -142,13 +167,7 @@ public ModuleManager(ModuleStore moduleStore, boolean local) {
...
         .compose(ares -> deleteCheckDep(id, ares))
-        .compose(res -> {
-          if (moduleStore == null) {
-            return Future.succeededFuture();
-          } else {
-            return moduleStore.delete(id).mapEmpty();
-          }
-        })
+        .compose(res -> moduleStore.delete(id).mapEmpty())
         .compose(res -> deleteInternal(id).mapEmpty());
```
julianladisch, line 171:
```suggestion
        .compose(res -> moduleStore.delete(id))
        .compose(res -> deleteInternal(id))
        .mapEmpty();
```

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -4530,17 +4008,133 @@ private void setupBasicModule(String tenant, String moduleId, String tenantPermi
...
+
+    api.createRestAssured3().given()
+        .header("Content-Type", "application/json")
+        .body(new JsonObject().put("id", okapiTenant).encode()).post("/_/proxy/tenants")
+        .then().statusCode(201);
+
+    List<ModuleDescriptor> modules = new LinkedList<>();
+    modules.add(new ModuleDescriptor("moduleA-1.0.0-SNAPSHOT.1"));
+    modules.add(new ModuleDescriptor("moduleA-1.0.0-SNAPSHOT.2"));
+    modules.add(new ModuleDescriptor("moduleA-1.0.0"));
```
julianladisch, line 4103:
```suggestion
    List<ModuleDescriptor> modules = List.of(
        new ModuleDescriptor("moduleA-1.0.0-SNAPSHOT.1"),
        new ModuleDescriptor("moduleA-1.0.0-SNAPSHOT.2"),
        new ModuleDescriptor("moduleA-1.0.0"));
```

okapi-core/src/test/java/org/folio/okapi/util/ModuleUtilTest.java
adamdickmeiss: ok

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
adamdickmeiss: ok

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
adamdickmeiss: ok

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
adamdickmeiss: ok

julianladisch (approved):
Looks good, thanks!

MikeTaylor:
I note that this facility is not documented in https://github.com/folio-org/okapi/blob/master/doc/guide.md

PR #1136, opened 2021-06-28
OKAPI-1025: /_/proxy/tenants/{tenant_id}/upgrade with body

julianladisch:
/_/proxy/tenants/{tenant_id}/upgrade upgrades each module to the latest version
and therefore doesn't needs a body to list the version to upgrade to.

Steps to Reproduce:

* POST /_/proxy/tenants/{tenant_id}/upgrade with body

Expected Results:

Error message about the useless body.

Actual Results:

Upgrades to latest versions ignoring the version specified in the body.

Additional Information:

The documentation in
https://github.com/folio-org/okapi/blob/v4.8.2/okapi-core/src/main/raml/okapi.raml
correctly explains that no body is needed and it upgrades to the latest version.
(The wording can be improved, though.)

Okapi provides
 * /_/proxy/tenants/{tenant_id}/install
 * /_/proxy/tenants/{tenant_id}/upgrade

being misleading at first glance because one might assume that install is intended
for initial install and upgrade is intended for upgrades.
Rejecting upgrade with body will prevent any unintentional results.

### DIFF
--- a/okapi-core/pom.xml
+++ b/okapi-core/pom.xml
@@ -153,7 +153,13 @@
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-inline</artifactId>
-      <version>3.4.6</version>
+      <version>3.11.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-junit-jupiter</artifactId>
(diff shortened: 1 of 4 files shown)


hjiebsco (approved):
Makes sense to me. @adamdickmeiss?

steveellis (approved):

PR #1139, opened 2021-08-06
OKAPI-1029: Cache CORS requests using Access-Control-Max-Age header

julianladisch:
Add a Access-Control-Max-Age header to allow browsers to cache CORS responses.
This cache is per URL only, though.
Without this header Chromium has a default cache time of 5 seconds, Firefox has no cache.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
@@ -6,7 +6,6 @@ import io.vertx.ext.web.Router;
 import io.vertx.ext.web.handler.CorsHandler;
 import java.util.List;
 import org.folio.okapi.bean.ModuleInstance;
-import org.folio.okapi.bean.Tenant;
 import org.folio.okapi.common.XOkapiHeaders;
 import org.folio.okapi.managers.TenantManager;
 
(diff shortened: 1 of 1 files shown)


steveellis (approved):
Seems like setting it for 2 hours will be a nice performance improvement by reducing the number of preflights.

okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
```diff
@@ -21,6 +20,7 @@
 
   public static final String DELEGATE_CORS = "delegate-CORS";
   public static final String DELEGATE_CORS_MODULE_INSTANCE = "delegate-CORS-module-instance";
+  private static final int ACCESS_CONTROL_MAX_AGE = 7200;
```
hjiebsco, line 23: Is there a specific reason to use 2 hours? BTW Spring default is 30 mins.

hjiebsco (approved):
Would using 30 mins be more consistent with other Spring based modules?

okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
julianladisch: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age
Chrome caps at 2 hours.

Okapi answers CORS OPTIONS pre-fetch requests, it does not pass it along to Spring based FOLIO modules.

For consistency across browsers I choose the maximum value that Chrome allows.

PR #1140, opened 2021-08-13
OKAPI-1031: "apt-get install okapi" should recreate /var/lib/okapi

julianladisch:
package removal removes /var/lib/okapi but not the user okapi.
Therefore a reinstall needs to create the directory even if the user already exists.


hjiebsco (approved):
LGTM. I have not used Okapi debian installation though.

adamdickmeiss (approved):

PR #1144, opened 2021-09-17
Supply module id for some tenant errors OKAPI-1034

adamdickmeiss:
Increased timeout for okapi-common's OkapiClient test.. which involves a DNS lookup that may not fail within 5 seconds (increasing to 10 seconds).


julianladisch (approved):
Looks good, thanks!

PR #1147, opened 2021-10-13
Disable X-Okapi-Trace header by default OKAPI-1038

adamdickmeiss:
This, strickly speaking, a breaking change since the default
behavior has changed. But nothing, except reporting tools,
will be affected by this. Headers can be controlled with
config trace_headers.


adamdickmeiss:
> LGTM. Would `enable_trace_headers` be more consistent naming than just `trace_headers`?

Ha.. I had that in my code, but then decided last minute to omit the enable_-prefix .. I thought it didn't add much value ..

PR #1148, opened 2021-10-15
Only put in token in cache if auth returns OK OKAPI-1037

adamdickmeiss:


steveellis:
The solution seems simple. If the problem is that non-200 stuff is getting added to the cache, we can just check for that.

When I saw the replication script I got very excited. With all the ways we have for testing things that involve multiple modules (rancher, karate, vagrant) the idea of being able to spin up a local build of everything and put in my own log statements, recompile in a few seconds and explore was very appealing. There's non-obvious stuff in the script like setting the env vars through the `/env` route. Even the fact that okapi will gracefully start (by posting the DeploymentDescriptor) and stop modules was surprising to me. I thought I would have to do `java -jar` myself on every module and got confused on what ports I should pick. Little did I know okapi would do that for me.

Sadly I wasn't able to make the script work on my machine though. After enabling mod-authtoken I get `No credentials match that login`. Yet I can see a row for `joe` in postgres. But posting the username and password of joe isn't enough for some reason to get a token, which makes the subsequent requests fail. Tried blowing away postgres and redoing things to no avail.

I see what you mean about the logic in `ProxyService.getNewIterator` but I have to admit I'm not as clear as I'd like to be about how things would go awry if those first two return statements aren't hit or what we're itererating on exactly here. More iterations are produced it seems but is that the source of things getting "through" with these "phase modules"?

I think I need a better understanding on what a handler is and what the phases are all about. I noticed that ModuleInstance has an isHandler property. Is that false modules like mod-aes?

hjiebsco (approved):
The fix makes sense and it works. Thanks.

steveellis (approved):
Was able to see it work using the repro script on a clean ubuntu vm.

Would love to know why the the user can't be found after enabling mod-authtoken, even though he's there in the db , but that's likely just a bug in my configuration on my mac.

adamdickmeiss:
> Was able to see it work using the repro script on a clean ubuntu vm.
> 
> Would love to know why the the user can't be found after enabling mod-authtoken, even though he's there in the db , but that's likely just a bug in my configuration on my mac.

This could be a bug that only occurs on Mac in general. Could be my script (unlikely :-) -- it could be in mod-login - in Okapi.. Remember in 99% of cases, Folio modules are execute in containers that are Linux based.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -478,8 +494,7 @@ private void sanitizeAuthHeaders(MultiMap headers) {
    */
   private void relayToRequest(HttpClientResponse res, ProxyContext pc,
                               ModuleInstance mi) {
-    if (XOkapiHeaders.FILTER_AUTH.equals(mi.getRoutingEntry().getPhase())
-        && res.headers().contains(XOkapiHeaders.MODULE_TOKENS)) {
```
steveellis, line 482: This seems right. We shouldn't be controlling things based on the presence or absence of these headers.

PR #1150, opened 2021-11-08
okapi reinstall OKAPI-1043

adamdickmeiss:
https://issues.folio.org/browse/OKAPI-1043

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
@@ -925,96 +925,98 @@ public class TenantManager implements Liveness {
       Map<String, ModuleDescriptor> modsEnabled, InstallJob job) {
 
     List<TenantModuleDescriptor> tml = job.getModules();
-    return DepResolution.installSimulate(modsAvailable, modsEnabled, tml).compose(res -> {
-      if (options.getSimulate()) {
-        return Future.succeededFuture(tml);
-      }
-      return jobs.add(t.getId(), job.getId(), job).compose(res2 -> {
-        Promise<List<TenantModuleDescriptor>> promise = Promise.promise();
-        Future<Void> future = Future.succeededFuture();
-        if (options.getAsync()) {
-          List<TenantModuleDescriptor> tml2 = new LinkedList<>();
-          for (TenantModuleDescriptor tm : tml) {
-            tml2.add(tm.cloneWithoutStage());
+    return DepResolution.installSimulate(modsAvailable, modsEnabled, tml, options.getReinstall())
+        .compose(res -> {
+          if (options.getSimulate()) {
+            return Future.succeededFuture(tml);
           }
-          promise.complete(tml2);
-        }
-        future = future.compose(x -> {
-          for (TenantModuleDescriptor tm : tml) {
-            tm.setStage(TenantModuleDescriptor.Stage.pending);
-          }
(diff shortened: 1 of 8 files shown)


ArgonV (approved):
This looks good on my end - but one thing I can't tell is what happens when deploy=false? Does Okapi just re-run the install API for the module with the tenant params?

adamdickmeiss:
> This looks good on my end - but one thing I can't tell is what happens when deploy=false? Does Okapi still re-run the install API for the module with the tenant init params?

Yes. deploy has no impact on this. The tenant init will of course fail if Okapi can not contact module , whether deployed by Okapi or elsewhere.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -925,96 +925,98 @@ private ModuleDescriptor findSystemInterface(Tenant tenant, String interfaceName
...
+                    logger.warn("Ignoring error for tenant {} module {}",
+                        t.getId(), tm.getId(), x.cause());
+                  }
+                  promise1.complete();
+                });
+                future = future.compose(x -> promise1.future());
+              } else {
+                future = future
+                    .compose(x -> installTenantModule(t, pc, options, modsAvailable,tm));
+              }
```
julianladisch, line 970:
```suggestion
              future = future.compose(x -> installTenantModule(t, pc, options, modsAvailable,tm));
              if (options.getIgnoreErrors()) {
                future = future.otherwise(e -> {
                  logger.warn("Ignoring error for tenant {} module {}", t.getId(), tm.getId(), e);
                  return null;
                });
              }
```

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -925,96 +925,98 @@ private ModuleDescriptor findSystemInterface(Tenant tenant, String interfaceName
...
-            tm.setStage(TenantModuleDescriptor.Stage.invoke);
-            return jobs.put(t.getId(), job.getId(), job);
-          });
-          if (options.getIgnoreErrors()) {
-            Promise<Void> promise1 = Promise.promise();
-            installTenantModule(t, pc, options, modsAvailable, tm).onComplete(x -> {
-              if (x.failed()) {
-                logger.warn("Ignoring error for tenant {} module {}",
-                    t.getId(), tm.getId(), x.cause());
+          return jobs.add(t.getId(), job.getId(), job).compose(res2 -> {
```
julianladisch, line 933:
```suggestion
          return jobs.add(t.getId(), job.getId(), job);
        }).compose(res -> {
```
Avoids nesting and reduced indentation.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: oh.. yes. Had a similar change here, but using recover. Using otherwise is probably more elegant.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: I don't see how this can work here.

adamdickmeiss: I've split it into two methods instead.

julianladisch (approved):
Looks good, thanks!

PR #1157, opened 2021-12-12
Hazelcast 4.2.2, logging log4j2 OKAPI-1048

adamdickmeiss:
Upgrade Hazelcast from 4.0.2 to 4.2.2 which is
what Vert.x 4.2.1 is using anyway. And set logging
type to log4j2 (rather than log4j).


julianladisch (approved):
Looks good, thanks!

PR #1168, opened 2021-12-15
OKAPI-1053: Add module users section to raml schema

MikhailFokanov:
As discussed (Jakub Skoczen, Vince Bareau, Craig McNally) module descriptor should have section for module user metadata. This section should contain user type and the permissions, that should be assigned for the user.

Acceptance criteria:

The section "user" is added to the raml schema
The section contains "type" and "permissions" property

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/bean/ModuleUser.java
+++ b/okapi-core/src/main/java/org/folio/okapi/bean/ModuleUser.java
@@ -1,0 +1,33 @@
+package org.folio.okapi.bean;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Tells how a module is to be deployed. Either by exec'ing a command (and
+ * killing the process afterwards), or by using command lines to start and stop
+ * it.
+ */
+@JsonInclude(Include.NON_NULL)
+public class ModuleUser {
+
+  private String type;
+  private String[] permissions;
+
+  public String getType() {
+    return type;
+  }
+
+  public void setType(String type) {
+    this.type = type;
+  }
+
+  public String[] getPermissions() {
+    return permissions;
+  }
+
+  public void setPermissions(String[] permissions) {
+    this.permissions = permissions;
+  }
+}
--- a/okapi-core/src/main/raml/ModuleDescriptor.json
+++ b/okapi-core/src/main/raml/ModuleDescriptor.json
@@ -83,7 +83,7 @@
     },
     "user" : {
       "description": "Module user descriptor",
-      "$ref": "LaunchDescriptor.json"
+      "$ref": "ModuleUser.json"
     }
   },
   "required": ["id"]
--- a/okapi-core/src/main/raml/ModuleDescriptor.json
+++ b/okapi-core/src/main/raml/ModuleDescriptor.json
@@ -80,6 +80,10 @@
     "launchDescriptor": {
       "description": "Default deployment for this module",
       "$ref": "LaunchDescriptor.json"
+    },
+    "user" : {
+      "description": "Module user descriptor",
+      "$ref": "LaunchDescriptor.json"
     }
   },
   "required": ["id"]
--- a/okapi-core/src/main/raml/ModuleUser.json
+++ b/okapi-core/src/main/raml/ModuleUser.json
@@ -2,7 +2,7 @@
   "$schema": "http://json-schema.org/draft-04/schema#",
   "id": "LaunchDescriptor.json",
   "title": "LaunchDescriptor",
-  "description": "Specifies how a module is deployed and undeployed",
+  "description": "Specifies the user that should be created for the module",
   "type": [ "object", "null" ],
   "additionalProperties" : false,
   "properties": {
--- a/okapi-core/src/main/raml/ModuleUser.json
+++ b/okapi-core/src/main/raml/ModuleUser.json
@@ -1,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "LaunchDescriptor.json",
+  "title": "LaunchDescriptor",
+  "description": "Specifies how a module is deployed and undeployed",
+  "type": [ "object", "null" ],
+  "additionalProperties" : false,
+  "properties": {
+    "type": {
+      "description": "The type of the module user",
+      "type": ["string"]
+    },
+    "permissions": {
+      "description": "Permission ID (usually module.service.method or similar)",
+      "type": ["array", "null"],
+      "items": {"type": "string"}
+    }
+  }
+}
+
--- a/okapi-core/src/main/raml/ModuleUser.json
+++ b/okapi-core/src/main/raml/ModuleUser.json
@@ -8,15 +8,15 @@
   "properties": {
     "type": {
       "description": "The type of the module user",
-      "type": ["string"]
+      "type": "string",
+      "enum": ["module", "tenant"],
+      "default": "module"
     },
     "permissions": {
       "description": "Permission ID (usually module.service.method or similar)",
-      "type": ["array", "null"],
+      "type": "array",
       "items": {
-        "type": "string",
-        "enum": ["module", "tenant"],
-        "default": "module"
+        "type": "string"
       }
     }
   }
--- a/okapi-core/src/main/raml/ModuleUser.json
+++ b/okapi-core/src/main/raml/ModuleUser.json
@@ -13,7 +13,11 @@
     "permissions": {
       "description": "Permission ID (usually module.service.method or similar)",
       "type": ["array", "null"],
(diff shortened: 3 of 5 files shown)


okapi-core/src/main/raml/ModuleDescriptor.json
```diff
@@ -80,6 +80,10 @@
     "launchDescriptor": {
       "description": "Default deployment for this module",
       "$ref": "LaunchDescriptor.json"
+    },
+    "user" : {
+      "description": "Module user descriptor",
+      "$ref": "LaunchDescriptor.json"
```
hjiebsco, line 86: Wrong `$ref`

okapi-core/src/main/raml/ModuleUser.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "LaunchDescriptor.json",
+  "title": "LaunchDescriptor",
+  "description": "Specifies how a module is deployed and undeployed",
```
hjiebsco, line 5: Copy/paste error. Does not match file name `ModuleUser.json`.

okapi-core/src/main/raml/ModuleDescriptor.json
MikhailFokanov-ebsco: oops. Fixed.

okapi-core/src/main/raml/ModuleUser.json
MikhailFokanov-ebsco: fixed

MikhailFokanov-ebsco: fixed

okapi-core/src/main/raml/ModuleUser.json
```diff
@@ -0,0 +1,20 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "LaunchDescriptor.json",
+  "title": "LaunchDescriptor",
+  "description": "Specifies the user that should be created for the module",
+  "type": [ "object", "null" ],
+  "additionalProperties" : false,
+  "properties": {
+    "type": {
+      "description": "The type of the module user",
+      "type": ["string"]
```
craigmcnally, line 11: Should this be an enum?

adamdickmeiss (changes requested):
I have no idea about the semantics of the new structure. When happy, be aware that you must create POJOs for it yourself in Okapi as code is NOT generated from schemas.  See ModuleDescriptpr.json. You'd probably also want a unit test where you post a module with this info to verify.

adamdickmeiss:
> As discussed (Jakub Skoczen, Vince Bareau, Craig McNally) module descriptor should have section for module user metadata. This section should contain user type and the permissions, that should be assigned for the user.
> 
> Acceptance criteria:
> 
> The section "user" is added to the raml schema The section contains "type" and "permissions" property

If the intention of this is *metadata* and NOT managed by Okapi , then why don't you use the "metadata" property of Okapi to hold this. This is "any" object and can hold whatever you like.

MikhailFokanov:
> > As discussed (Jakub Skoczen, Vince Bareau, Craig McNally) module descriptor should have section for module user metadata. This section should contain user type and the permissions, that should be assigned for the user.
> > Acceptance criteria:
> > The section "user" is added to the raml schema The section contains "type" and "permissions" property
> 
> If the intention of this is _metadata_ and NOT managed by Okapi , then why don't you use the "metadata" property of ModuleDescriptor to hold this. This is "any" object and can hold whatever you like.

@craigmcnally do you like the idea of adding "user" to "metadata" property?

craigmcnally:
> > > As discussed (Jakub Skoczen, Vince Bareau, Craig McNally) module descriptor should have section for module user metadata. This section should contain user type and the permissions, that should be assigned for the user.
> > > Acceptance criteria:
> > > The section "user" is added to the raml schema The section contains "type" and "permissions" property
> > 
> > 
> > If the intention of this is _metadata_ and NOT managed by Okapi , then why don't you use the "metadata" property of ModuleDescriptor to hold this. This is "any" object and can hold whatever you like.
> 
> @craigmcnally do you like the idea of adding "user" to "metadata" property?

Would this mean that it couldn't be structured / controlled by a schema?  I see metadata is just "object" w/o a schema.

craigmcnally:
@adamdickmeiss is there any precedent for how metadata is used in the module descriptor?  Is it currently being used for anything?

adamdickmeiss:
> @adamdickmeiss is there any precedent for how metadata is used in the module descriptor? Is it currently being used for anything?

AFAIK no. Of course, the drawback is that the structure is undefined, so if you wish some structure it's no good.

craigmcnally:
IMO having a well defined structure for this is desirable, possibly including an enumeration of valid values (I can probably be dissuaded of that part though)

okapi-core/src/main/raml/ModuleUser.json
```diff
@@ -0,0 +1,24 @@
...
+    },
+    "permissions": {
+      "description": "Permission ID (usually module.service.method or similar)",
+      "type": ["array", "null"],
+      "items": {
+        "type": "string",
+        "enum": ["module", "tenant"],
+        "default": "module"
+      }
```
adamdickmeiss, line 20: Why is this an enum and not the "type" earlier.. I thought the purpose of this array was to list the permissions needed for that user. If we don't, how do we know what the module needs?

okapi-core/src/main/raml/ModuleUser.json
adamdickmeiss: Yes. But not yet.

adamdickmeiss (changes requested):

okapi-core/src/main/raml/ModuleDescriptor.json
```diff
@@ -80,6 +80,10 @@
     "launchDescriptor": {
       "description": "Default deployment for this module",
       "$ref": "LaunchDescriptor.json"
+    },
+    "user" : {
```
craigmcnally, line 84: Do we foresee the need to ever define more than one module user?

adamdickmeiss: How would those be distinguished from one another? This is per module it seems.

okapi-core/src/main/raml/ModuleUser.json
```diff
@@ -0,0 +1,24 @@
...
+  "additionalProperties" : false,
+  "properties": {
+    "type": {
+      "description": "The type of the module user",
+      "type": "string",
+      "enum": ["module", "tenant"],
+      "default": "module"
+    },
+    "permissions": {
+      "description": "Permission ID (usually module.service.method or similar)",
```
adamdickmeiss, line 16: I don't understand this description. Isn't this the permissions that the module user requires to operate?

adamdickmeiss (approved):

okapi-core/src/main/java/org/folio/okapi/bean/ModuleUser.java
```diff
@@ -0,0 +1,33 @@
+package org.folio.okapi.bean;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Tells how a module is to be deployed. Either by exec'ing a command (and
+ * killing the process afterwards), or by using command lines to start and stop
+ * it.
```
julianladisch, line 10: Copy & paste from https://github.com/folio-org/okapi/blob/v4.12.0/okapi-core/src/main/java/org/folio/okapi/bean/LaunchDescriptor.java , needs changes for ModuleUser.

okapi-core/src/main/raml/ModuleDescriptor.json
julianladisch: I also suggest multiple module users with different permissions. For example a migration might require more permissions than the regular work.
We can easily add some identifier to distinguish them.

MikhailFokanov: The additional deployment complexity of managing identifiers for users and providing of multiples usernames along with credentials to the module environment variables without corresponding business requirements seems to be a premature optimization. It is also inconvenient to provide Map<String,String> of arbitrary size (username and pass for each user) in environment variable.

okapi-core/src/main/java/org/folio/okapi/bean/ModuleUser.java
MikhailFokanov: done

okapi-core/src/main/raml/ModuleUser.json
MikhailFokanov: done

okapi-core/src/main/raml/ModuleUser.json
MikhailFokanov: done

adamdickmeiss (changes requested):
Close this without merge?

PR #1169, opened 2021-12-16
OKAPI-1054: WebClientFactory to avoid socket leaks (okapi-common)

julianladisch:


okapi-common/src/main/java/org/folio/okapi/common/WebClientFactory.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Factory for WebClient instances to avoid web socket leaks.
+ */
+public final class WebClientFactory {
+  private static final Map<Vertx, WebClient> clients = new ConcurrentHashMap<>();
+
+  private WebClientFactory() {
+    throw new UnsupportedOperationException("Utility classes cannot be instantiated");
+  }
```
adamdickmeiss, line 16: Not tested. okapi-testing has a UtilityClassTester.

okapi-common/src/main/java/org/folio/okapi/common/WebClientFactory.java
```diff
@@ -0,0 +1,25 @@
...
+
+  /**
+   * Get a WebClient, returns the same instance for the same Vertx instance.
+   */
+  public static WebClient getWebClient(Vertx vertx) {
+    return clients.computeIfAbsent(vertx, x -> WebClient.create(vertx));
+  }
+}
```
adamdickmeiss, line 24: In some cases, modules pass WebClientOptions to be used. I think they can only be used on construction, so perhaps adding a constructor for that?

okapi-common/src/main/java/org/folio/okapi/common/WebClientFactory.java
julianladisch: Yes, thanks!

okapi-common/src/main/java/org/folio/okapi/common/WebClientFactory.java
julianladisch: Added.

adamdickmeiss (approved):
Nice

PR #1175, opened 2021-12-19
OKAPI-1058: Reject MDC lookups mitigating log4j vuln (CVE-2021-45105)

julianladisch:
If Okapi filters out strings that contain `${` it prevents any MDC lookup in the modules that haven't upgraded to log4j >= 2.17.0 and mitigates the log4j issue (CVE-2021-45105).

No modules need to be updated to log4j 2.17.0 if Okapi filters out invalid values that triggers the denial-of-service bug.

`${` is the default prefix for MDC lookups and used in FOLIO: https://github.com/apache/logging-log4j2/blob/rel/2.13.3/log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/StrSubstitutor.java#L146-L151

MDC is used for these values:

* tenant id
* request id
* module id
* user id

A valid tenant id cannot contain `${`: https://github.com/folio-org/okapi/blob/v4.11.0/okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java#L739
A valid request id cannot contain `${`: https://github.com/folio-org/okapi/blob/v4.11.0/okapi-core/src/main/java/org/folio/okapi/util/ProxyContext.java#L78-L79
A valid module id cannot contain `${` because no product name contains it and the semvar part doesn't contain it.
A valid user id is a UUID and doesn't contain `${`.

Therefore rejecting `${` only rejects invalid values.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -541,29 +541,26 @@ public class ProxyService {
     // it will get read into a buffer somewhere.
 
     ProxyContext pc = new ProxyContext(ctx, waitMs);
+    final MultiMap headers = ctx.request().headers();
 
     // It would be nice to pass the request-id to the client, so it knows what
     // to look for in Okapi logs. But that breaks the schemas, and RMB-based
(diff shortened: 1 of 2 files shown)


adamdickmeiss (approved):
ok

PR #1178, opened 2021-12-21
Set JAVA_HOME when building deb package FOLIO-3371

adamdickmeiss:
https://issues.folio.org/browse/FOLIO-3371
The build process can be tested "by hand" by doing `fakeroot debian/rules build` .


julianladisch (approved):
Debian has a policy that programs should run without envs set.

Therefore approve this fix.

Looks good, thanks!

PR #1181, opened 2022-01-08
Install reports all errors OKAPI-662

adamdickmeiss:
https://issues.folio.org/browse/OKAPI-662

Current /install endpoint only reports the first incompatible version.
This change reports all issues in one shot.


okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -233,8 +233,8 @@ public void setProxyService(ProxyService px) {
       mods.put(md.getId(), md);
     }
     if (modTo == null) {
-      String deps = DepResolution.checkAllDependencies(mods);
-      if (!deps.isEmpty()) {
+      List<String> errors = DepResolution.checkEnabled(mods);
+      if (!errors.isEmpty()) {
         return Future.succeededFuture(); // failures even before we remove a module
```
julianladisch, line 238: You haven't changed this, but why does this return success suppressing both any existing and any new resolution errors?

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -28,6 +29,135 @@ private DepResolution() {
...
+        product = md.getProduct();
+        no = 0;
+      } else if (no >= limit) {
+        it.remove();
+      }
+      no++;
+    }
+  }
+
+  static boolean moduleDepProvided(List<ModuleDescriptor> modules, ModuleDescriptor md) {
```
julianladisch, line 122: Is it intended that optional dependencies are checked is if they were required dependencies? If yes can this added to the javadoc?

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -28,6 +29,135 @@ private DepResolution() {
...
+      while (iterator.hasNext()) {
+        ModuleDescriptor md = iterator.next();
+        if (moduleDepProvided(result, md)) {
+          result.add(md);
+          iterator.remove();
+          more = true;
+        }
+      }
+    }
+    // add remaining result to modules (such as modules with optional interfaces requirements
```
julianladisch, line 157: Multiple modules with unsatisfied optional dependency can be sorted correctly if all optional dependencies that cannot satisfied are removed.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -143,130 +273,348 @@ private static String checkOneDependency(ModuleDescriptor md, InterfaceDescripto
...
+            .collect(Collectors.joining(", ")));
+
+    logger.info("Input install list {}", () ->
+        tml.stream()
+            .map(TenantModuleDescriptor::getId)
+            .collect(Collectors.joining(", ")));
+
+    List<TenantModuleDescriptor> tml2 = new ArrayList<>();
+
+    // go through disabled modules
```
julianladisch, line 323:
```suggestion
    // go through modules that need to be disabled
```

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -143,130 +273,348 @@ private static String checkOneDependency(ModuleDescriptor md, InterfaceDescripto
...
+          }
         }
       }
     }
-    if (list.isEmpty()) {
-      return "";
-    } else {
-      return String.join("\n", list);
+
+    // go through enabled/updated modules
```
julianladisch, line 340:
```suggestion
    // go through modules that need to be enabled/updated
```

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -143,130 +273,348 @@ private static String checkOneDependency(ModuleDescriptor md, InterfaceDescripto
...
+            .collect(Collectors.joining(", ")));
+
+    List<TenantModuleDescriptor> tml2 = new ArrayList<>();
+
+    // go through disabled modules
+    Iterator<ModuleDescriptor> moduleIterator = sortedList.descendingIterator();
+    while (moduleIterator.hasNext()) {
+      ModuleDescriptor md = moduleIterator.next();
+      Iterator<TenantModuleDescriptor> iterator = tml.iterator();
+      while (iterator.hasNext()) {
```
julianladisch, line 328: This nested loop could be replaced by a hash lookup. Simply set up a `Set<String> disable = new HashSet<>()` similar to the existing added HashSet.
The same below for enable.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: The problem is that if a set of modules are already inconsistent, then we allow it to proceed anyway.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: Could be done.. I tried, but not so easy.. The same module could be mentioned twice.. That could be the case for enable, uptodate actions..

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: ok

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: Now done, but not by removal, but by checking all interfaces provided (if not part of that, the optional interface is ok)

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: Added javadoc and changed the method a bit.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
julianladisch: Can we have a warning in the log?

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
```diff
@@ -190,443 +174,715 @@ public void testLatest(TestContext context) {
...
+  }
+
+  @Test
+  public void testUpgrade1() {
+    List<TenantModuleDescriptor> tml = enableList(mdD200);
+    DepResolution.install(map(mdA100, mdB, mdC, mdA110, mdA200, mdD100, mdD200), map(mdB), tml, false);
+    Assert.assertEquals(Json.encodePrettily(tml), 3, tml.size());
+    assertDisable(tml, 0, mdB);
+    assertEnable(tml, 1, mdA200);
+    assertEnable(tml, 2, mdD200);
```
julianladisch, line 251:
```suggestion
    assertThat(tml, contains(enable(mdD200), disable(mdB), enable(mdA200)));
```
Matchers.contains checks size, order and each element.

Requires
```
import static org.hamcrest.MatcherAssert.assertThat;
import static org.hamcrest.Matchers.contains;
```
and
```
import org.hamcrest.Description;
import org.hamcrest.TypeSafeMatcher;

public static class TenantModuleDescriptorMatcher extends TypeSafeMatcher<TenantModuleDescriptor> {

  private final Action action;
  private final ModuleDescriptor md;
  private final ModuleDescriptor from;

  public TenantModuleDescriptorMatcher(Action action, ModuleDescriptor md, ModuleDescriptor from) {
      this.action = action;
      this.md = md;
      this.from = from;
  }

  @Override
  public void describeTo(final Description description) {
      description.appendText("matches TenantModuleDescriptor(action = " + action + ", id = " + md.getId()
        + ", from = " + (from == null ? null : from.getId()) + ")");
  }

  @Override
  public boolean matchesSafely(final TenantModuleDescriptor tm) {
    return action == tm.getAction() && md.getId() == tm.getId()
        && (from == null ? tm.getFrom() == null : tm.getFrom() == from.getId());
  }

  public static TenantModuleDescriptorMatcher enable(ModuleDescriptor md) {
    return upgrade(md, null);
  }

  public static TenantModuleDescriptorMatcher upgrade(ModuleDescriptor md, ModuleDescriptor from) {
    return new TenantModuleDescriptorMatcher(Action.enable, md, from);
  }

  public static TenantModuleDescriptorMatcher disable(ModuleDescriptor md) {
    return new TenantModuleDescriptorMatcher(Action.disable, md, null);
  }

  public static TenantModuleDescriptorMatcher upToDate(ModuleDescriptor md) {
    return new TenantModuleDescriptorMatcher(Action.uptodate, md, null);
  }
}
```
and `import static` of TenantModuleDescriptorMatcher.*

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -28,6 +29,160 @@ private DepResolution() {
...
+   */
+  public static void install(
+      Map<String, ModuleDescriptor> modsAvailable,
+      Map<String, ModuleDescriptor> modsEnabled, List<TenantModuleDescriptor> tml,
+      boolean reinstall) {
+    installMaxIterations(modsAvailable, modsEnabled, tml, reinstall, 100);
+  }
+
+  /**
+   * Return top-N set of modules - in order of module ID.
```
julianladisch, line 106:
```suggestion
   * Sort mdl in descending order and remove all but the top-N for each product.
```

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -28,6 +29,160 @@ private DepResolution() {
...
+   * @param allProvided interfaces for all modules
+   * @param md          module to check.
+   * @return            true if interface requirements are met
+   */
+  static boolean moduleDepProvided(
+      List<ModuleDescriptor> modules, Set<String> allProvided,
+      ModuleDescriptor md) {
+
+    for (InterfaceDescriptor req : md.getRequiresOptionalList()) {
+      if (allProvided.contains(req.getId())) {
```
julianladisch, line 141: Skipping optional interfaces that are not in allProvided is correct.
Is skipping required interfaces that are not in allProvided correct?

adamdickmeiss: Yes. In this case it is.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -143,130 +299,339 @@ private static String checkOneDependency(ModuleDescriptor md, InterfaceDescripto
...
-   * @return empty string if all OK; error message otherwise
-   */
-  public static String checkAllDependencies(Map<String, ModuleDescriptor> available) {
-    Collection<ModuleDescriptor> testList = available.values();
-    return checkDependencies(testList, testList, false);
+  private static void addTenantModule(
+      List<TenantModuleDescriptor> tml, String id, String from,
+      TenantModuleDescriptor.Action action) {
+    if (from != null && action == TenantModuleDescriptor.Action.enable && !id.equals(from)) {
+      if (!new ModuleId(id).getProduct().equals(new ModuleId(from).getProduct())) {
```
julianladisch, line 389: The 2 ifs can be merged.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -143,130 +299,339 @@ private static String checkOneDependency(ModuleDescriptor md, InterfaceDescripto
...
-
-    Iterator<TenantModuleDescriptor> it = tml.iterator();
-    TenantModuleDescriptor tm;
-    while (it.hasNext()) {
-      tm = it.next();
-      TenantModuleDescriptor.Action action = tm.getAction();
-      String id = tm.getId();
-      if (logger.isInfoEnabled()) {
-        logger.info("getNextTM: loop id {} action {}", id, action.name());
+  private static boolean checkMultiple(
```
julianladisch, line 447:
```suggestion
  /**
   * For each entry in providedInterfaces with more than one ModuleInterface add an error message to errors.
   * If fix is true and it is not listed in stickyModules also disable it by adding an disable entry to tml and
   * by removing it from modsEnabled.
   */
  private static boolean checkMultiple(
```

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -143,130 +299,339 @@ private static String checkOneDependency(ModuleDescriptor md, InterfaceDescripto
...
+
+  private static boolean checkRequired(
+      Map<String, ModuleDescriptor> modsAvailable,
+      Map<String, ModuleDescriptor> modsEnabled, List<TenantModuleDescriptor> tml,
+      boolean fix, List<String> errors, Map<String, List<ModuleInterface>> providedInterfaces,
+      Map<String, List<ModuleInterface>> requiredInterfaces, Set<String> stickyModules) {
+
+    for (Map.Entry<String, List<ModuleInterface>> entry : requiredInterfaces.entrySet()) {
+      List<ModuleInterface> providedModuleInterfaces = providedInterfaces.get(entry.getKey());
+      if (providedModuleInterfaces == null) {
```
julianladisch, line 484:
```
      if (providedInterfaces.get(entry.getKey() != null) {
        continue;  // required interfaces already provided, nothing to do
      }
      logger.info("Interface {} undefined and required", entry.getKey());
```
To reduce indentation and for better readability.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -143,130 +299,339 @@ private static String checkOneDependency(ModuleDescriptor md, InterfaceDescripto
...
+                stickyModules.add(mdFound.getId());
+              } else {
+                logger.info("Removing {}", id);
+                modsEnabled.remove(id);
+                addTenantModule(tml, id, null, TenantModuleDescriptor.Action.disable);
+              }
+              return true;
+            }
+          }
+          errors.add(messages.getMessage("10211", entry.getKey(), req.moduleDescriptor.getId()));
```
julianladisch, line 510:
```suggestion
          if (! fix) {
            errors.add(messages.getMessage("10211", entry.getKey(), req.moduleDescriptor.getId()));
            continue;
          }
          Map<String, ModuleDescriptor> modules =
              checkInterfaceDepAvailable(modsAvailable, req.interfaceDescriptor);
          if (modules.isEmpty()) {
            errors.add(messages.getMessage("10211", entry.getKey(), req.moduleDescriptor.getId()));
            continue;
          }
          if (modules.size() > 1) {
            errors.add(messages.getMessage("10210", entry.getKey(), req.moduleDescriptor.getId(),
                String.join(", ", modules.keySet())));
            return false;
          }
          ModuleDescriptor mdFound = modules.values().iterator().next();
          String id = req.moduleDescriptor.getId();
          if (stickyModules.contains(req.moduleDescriptor.getId())) {
            logger.info("Enable {}", mdFound.getId());
            modsEnabled.put(mdFound.getId(), mdFound);
            addTenantModule(tml, mdFound.getId(), null, TenantModuleDescriptor.Action.enable);
            stickyModules.add(mdFound.getId());
          } else {
            logger.info("Removing {}", id);
            modsEnabled.remove(id);
            addTenantModule(tml, id, null, TenantModuleDescriptor.Action.disable);
          }
          return true;
```
Guard clauses for less indentation, less nested ifs and more readability.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -143,130 +299,339 @@ private static String checkOneDependency(ModuleDescriptor md, InterfaceDescripto
...
+        }
       }
-      if (action == TenantModuleDescriptor.Action.disable && modsEnabled.containsKey(id)) {
-        logger.info("getNextTM: return tm for action disable");
-        return tm;
+    }
+    return false;
+  }
+
+  private static boolean checkCompatible(
```
julianladisch, line 517: Can you add javadoc explaining the return value?

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -143,130 +299,339 @@ private static String checkOneDependency(ModuleDescriptor md, InterfaceDescripto
...
-          tm.setAction(TenantModuleDescriptor.Action.uptodate);
+        } else {
+          if (modsEnabled.containsKey(id) && !reinstall) {
+            tm.setAction(TenantModuleDescriptor.Action.uptodate);
+          } else {
+            String product = new ModuleId(id).getProduct();
+            for (String enabledId : modsEnabled.keySet()) {
+              String enabledProduct = new ModuleId(enabledId).getProduct();
+              if (enabledProduct.equals(product)) {
+                if (stickyModules.contains(enabledId) && !id.equals(enabledId)) {
```
julianladisch, line 625: Can you reduce the complexity, here we have this nesting: for - if - if-else - if-else - for - if - if. I can hardly follow what happens in these 7 level of nesting.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -402,247 +691,33 @@ private static TenantModuleDescriptor getNextTM(Map<String, ModuleDescriptor> mo
...
-  }
-
-  private static void upgradeLeafs(
-      Map<String, ModuleDescriptor> modsAvailable,
-      Map<String, ModuleDescriptor> modsEnabled, List<TenantModuleDescriptor> tml) {
-    while (upgradeLeafs2(modsAvailable, modsEnabled, tml)) {
-      // something upgraded.. try again
-    }
-  }
+  static Map<String, ModuleDescriptor> checkInterfaceProvAvailable(
```
julianladisch, line 694: Can you add javadoc about the return value?

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
```diff
@@ -637,426 +898,460 @@ public void testCheckDependencies() {
...
+    DepResolution.install(map(mdA100, mdA110, mdE100, mdE110), map(), tml, false);
+    Assert.assertEquals(Json.encodePrettily(tml), 2, tml.size());
+    assertEnable(tml, 0, mdA100);
+    assertEnable(tml, 1, mdE100);
   }
 
+  @Test
+  public void testCheckInterfaceDepAvailable() {
+    Map<String, ModuleDescriptor> available = new TreeMap<>();
+    available.put(mdA100.getId(), mdA110);
```
julianladisch, line 1294:
```suggestion
    available.put(mdA110.getId(), mdA110);
```

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
```diff
@@ -637,426 +898,460 @@ public void testCheckDependencies() {
...
+    available.put(mdA111.getId(), mdA111);
+    available.put(mdA100.getId(), mdA100);
+    available.put(mdB.getId(), mdB);
+    InterfaceDescriptor req = new InterfaceDescriptor("int", "1.0");
+    Map<String,ModuleDescriptor> products = DepResolution.checkInterfaceDepAvailable(available, req);
+    Assert.assertEquals(2, products.size());
+    Assert.assertEquals(mdA111, products.get(mdA111.getProduct()));
+    Assert.assertEquals(mdB, products.get(mdB.getProduct()));
```
julianladisch, line 1302:
```suggestion
    assertThat(products, is(Map.of(mdA111.getProduct(), mdA111, mdB.getProduct(), mdB)));
```

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -28,6 +29,160 @@ private DepResolution() {
...
+        no = 0;
+      } else if (no >= limit) {
+        it.remove();
+      }
+      no++;
+    }
+  }
+
+  /**
+   * Check if module's required/optional interfaces are provided by set of modules.
```
julianladisch, line 129:
```suggestion
   * Check if md's required and optional interfaces are provided by set of modules.
   * The interface version is not checked, only the interface name.
   * Only interfaces listed in allProvided are checked, this allows to exclude some optional
   * interfaces from the check.
```

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: Sure

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: sure

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: yep

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: done

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -143,130 +299,339 @@ private static String checkOneDependency(ModuleDescriptor md, InterfaceDescripto
...
+        errors.add(messages.getMessage("10213", modules, entry.getKey()));
       }
-      if (action == TenantModuleDescriptor.Action.enable && !modsEnabled.containsKey(id)) {
-        logger.info("getNextMT: return tm for action enable");
-        return tm;
+    }
+    return false;
+  }
+
+  private static boolean checkRequired(
```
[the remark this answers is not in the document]
adamdickmeiss: done

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: I've partially done this.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: done

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: yep

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
adamdickmeiss: ok

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
adamdickmeiss: Added. Thx

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
adamdickmeiss: ok

okapi-core/src/test/java/org/folio/okapi/util/DepResolutionTest.java
```diff
@@ -190,443 +160,606 @@ public void testLatest(TestContext context) {
 
     DepResolution.getLatestProducts(2, mdl);
 
-    context.assertEquals(5, mdl.size());
-    context.assertEquals(mdE100, mdl.get(0));
-    context.assertEquals(mdC, mdl.get(1));
-    context.assertEquals(mdB, mdl.get(2));
-    context.assertEquals(mdA200, mdl.get(3));
-    context.assertEquals(mdA110, mdl.get(4));
+    Assert.assertEquals(5, mdl.size());
```
julianladisch, line 163: `Matchers.contains` also checks the size, no need to do it here.

julianladisch (approved):
Looks good, thanks!

PR #1182, opened 2022-01-13
npmSnapshot/preRelease only filter OKAPI-1065

adamdickmeiss:
npmSnapshot/preRelease was previously a boolean, now an enum with
values "false", "true", "only". The fist 2 works exactly as before.
The new "only" filters by npmSnapshot/preRelease only.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/util/ModuleUtil.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/ModuleUtil.java
@@ -27,8 +27,8 @@ public final class ModuleUtil {
     TenantInstallOptions options = new TenantInstallOptions();
 
     options.setSimulate(getParamBoolean(params, "simulate", false));
-    options.setPreRelease(getParamBoolean(params, "preRelease", true));
-    options.setNpmSnapshot(getParamBoolean(params, "npmSnapshot", true));
+    options.setPreRelease(getParamVersionFilter(params, "preRelease"));
+    options.setNpmSnapshot(getParamVersionFilter(params, "npmSnapshot"));
     options.setDeploy(getParamBoolean(params, "deploy", false));
     options.setPurge(getParamBoolean(params, "purge", false));
     options.setTenantParameters(params.get("tenantParameters"));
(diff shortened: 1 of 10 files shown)


okapi-core/src/main/java/org/folio/okapi/util/ModuleUtil.java
```diff
@@ -39,6 +40,24 @@ public static TenantInstallOptions createTenantOptions(MultiMap params) {
     return options;
   }
 
+  /**
+   * Lookup boolean preRelease/npmSnapshot parameter in HTTP request.
+   * @param params HTTP server request parameters
+   * @param name name of query parameter
+   * @return boolean value (possibly empty)
```
julianladisch, line 47:
```suggestion
   * @return returns false for "false", true for "only", and Optional.empty() for "true" and null/undefined
   * @throws DecodeException on invalid value
```

okapi-core/src/main/java/org/folio/okapi/util/ModuleUtil.java
```diff
@@ -39,6 +40,24 @@ public static TenantInstallOptions createTenantOptions(MultiMap params) {
...
+  public static Optional<Boolean> getParamVersionFilter(MultiMap params, String name) {
+    String v = params.get(name);
+    if (v == null || "true".equals(v)) {
+      return Optional.empty();
+    } else if ("false".equals(v)) {
+      return Optional.of(false);
+    } else if ("only".equals(v)) {
+      return Optional.of(true);
+    }
+    throw new DecodeException("Bad boolean for parameter " + name + ": " + v);
```
julianladisch, line 58:
```suggestion
    throw new DecodeException("Expected "true", "false", "only" or undefined/null for parameter " + name + ", but got: " + v);
```

okapi-core/src/main/raml/okapi.raml
```diff
@@ -373,12 +373,12 @@ types:
         default: true
       preRelease:
         description: Whether to allow pre-release modules in dependency check
-        type: boolean
+        enum: ["true","false","only"]
         required: false
         default: true
```
julianladisch, line 378:
```suggestion
        default: "true"
```
All strings should be quoted to make the RAML linter happy.

dcrossleyau:
There are 9 more cases where `default: ...` needs to be quoted string.

adamdickmeiss:
I guess I should see if I can avoid repeating those query parameters and have them defined once in the RAML.

julianladisch (approved):
Looks good, thanks!

PR #1183, opened 2022-01-14
Create class for snapshot filter handling OKAPI-1065

adamdickmeiss:
This is a follow-up to OKAPI-1065. While this is more lines of code,
the logic is now all encapsulated in one class, which makes it clearer
how this works and easier to maintain.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/util/ModuleVersionFilter.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/ModuleVersionFilter.java
@@ -23,11 +23,8 @@ public class ModuleVersionFilter {
    * @param params HTTP parameters
    */
   public ModuleVersionFilter(MultiMap params) {
-    Optional<Boolean> o = getParamVersionFilter(params, "npmSnapshot");
-    npmSnapshot = o.isEmpty() ? null : o.get();
-
-    o = getParamVersionFilter(params, "preRelease");
-    preRelease = o.isEmpty() ? null : o.get();
+    npmSnapshot = getParamVersionFilter(params, "npmSnapshot").orElse(null);
+    preRelease = getParamVersionFilter(params, "preRelease").orElse(null);
   }
 
   /**
--- a/okapi-core/src/main/java/org/folio/okapi/util/ModuleVersionFilter.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/ModuleVersionFilter.java
@@ -1,0 +1,71 @@
+package org.folio.okapi.util;
+
+import io.vertx.core.MultiMap;
+import io.vertx.core.json.DecodeException;
+import java.util.Optional;
+import org.folio.okapi.common.ModuleId;
+
+public class ModuleVersionFilter {
+  Boolean npmSnapshot;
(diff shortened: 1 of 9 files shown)


okapi-core/src/main/java/org/folio/okapi/util/ModuleVersionFilter.java
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Construct from HTTP query parameters "npmSnapshot", "preRelease".
+   * @param params HTTP parameters
+   */
+  public ModuleVersionFilter(MultiMap params) {
+    Optional<Boolean> o = getParamVersionFilter(params, "npmSnapshot");
+    npmSnapshot = o.isEmpty() ? null : o.get();
+
+    o = getParamVersionFilter(params, "preRelease");
+    preRelease = o.isEmpty() ? null : o.get();
```
julianladisch, line 30:
```suggestion
    npmSnapshot = getParamVersionFilter(params, "npmSnapshot").orElse(null);
    preRelease = getParamVersionFilter(params, "preRelease").orElse(null);
```
Or rewrite getParamVersionFilter to return a Boolean and not an `Optional<Boolean>`.

okapi-core/src/main/java/org/folio/okapi/util/ModuleVersionFilter.java
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Lookup boolean preRelease/npmSnapshot parameter in HTTP request.
+   * @param params HTTP server request parameters
+   * @param name name of query parameter
+   * @return returns false for "false", true for "only", and
+   *     Optional.empty() for "true" and null/undefined
+   * @throws DecodeException on invalid value
+   */
+  static Optional<Boolean> getParamVersionFilter(MultiMap params, String name) {
```
julianladisch, line 57: Can you also replace `Optional<Boolean>` by `Boolean` here?

adamdickmeiss: ok. I'll try.. Just note that SonarQube warned when I returned null earlier..

adamdickmeiss: I tried it again #1184. So I'll keep the  optional and use orElse.

julianladisch: SonarQube's warning is a false positive and should be suppressed. The method is used in this class only and the value is intended to be one of three states, one of them is null. There is no danger that someone forgets that the value might be null.

Using `Optional<Boolean> v` with `v.get()` may throw NPE, using `Boolean v` and accessing `v` may also throw NPE. This is exactly the same danger.

If you consider SonarQube's warning about the method to be valid then this assignment to a Boolean
`npmSnapshot = getParamVersionFilter(params, "npmSnapshot").orElse(null)`
is equally dangerous.

julianladisch (approved):
Looks good, thanks!
I don't agree on the Optional but approve because it's a minor issue.

PR #1186, opened 2022-01-16
install with action=enable and purge OKAPI-1064

adamdickmeiss:
Enable with purge will make two calls to the module to be enabled.
The first being disable (as if the module was disabled). Errors
are ignored for this first operation as some modules return 400
if a module was not enabled already. The second will proceed as
usual.


okapi-core/src/test/java/org/folio/okapi/InstallTest.java
```diff
@@ -1092,17 +1096,22 @@ void deployAsyncInitModule(TestContext context, String module, int port) {
...
+        ).encode())
+        .post("/_/proxy/tenants/" + tenant + "/install" + installParameters)
         .then().statusCode(201)
-        .body(equalTo("[ {" + LS
-            + "  \"id\" : \""+ module + "\"," + LS
-            + "  \"action\" : \"enable\"" + LS
-            + "} ]"))
+        .body("[0].id", is(module))
+        .body("[0].action", is(action))
         .extract().response();
```
julianladisch, line 1115: `.extract().header("Location");` avoids `Response r` variable.

adamdickmeiss: ok.. If you look for things in Okapi test that could be improved you'll find a lot and I could be spending quite some time on this ...

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -344,8 +344,33 @@ private void waitTenantInit(Tenant tenant, ModuleInstance getInstance,
...
-                                             ModuleDescriptor mdFrom, ModuleDescriptor mdTo,
-                                             ProxyContext pc) {
+      ModuleDescriptor mdFrom, ModuleDescriptor mdTo,
+      ProxyContext pc) {
+
+    Future<Void> future;
+    if (mdTo != null && options.getPurge()) {
+      // enable with purge is turned into purge on its own, followed by regular enable
+      mdFrom = null;
+      future = invokeTenantInterface1(tenant, options, mdTo, null, pc)
```
julianladisch, line 354: Is it intended that not the old module (mdFrom) but the new module (mdTo) executes the purge?

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -344,8 +344,33 @@ private void waitTenantInit(Tenant tenant, ModuleInstance getInstance,
...
+            return null;
+          })
+          .map(x -> {
+            options.setPurge(false);
+            return null;
+          });
+    } else {
+      future = Future.succeededFuture();
+    }
+    final ModuleDescriptor mdFromFinal = mdFrom;
```
julianladisch, line 367: Another possibility is to pass the mdFromFinal as Future success value.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: Yes

adamdickmeiss: The from is irrelevant and discarded because the idea is to reset the module to that version initialized (mdTo .. id property in install). Suppose the module was already running in some other version (according to Okapi), that information is explicitly reset.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
adamdickmeiss: ok

julianladisch (approved):
Looks good, thanks!

PR #1194, opened 2022-01-26
Check permission names during install OKAPI-1070

adamdickmeiss:
Check list of modules that are present during end of install operation. At this point, the
result of the check is only logged as there are many undefined permissions in FOLIO at the
moment.

See https://issues.folio.org/browse/FOLIO-3435 and attachment for how this is reported for FOLIO modules.


okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -657,8 +659,14 @@ static void installMaxIterations(
     if (!errors.isEmpty()) {
       throw new OkapiError(ErrorType.USER, String.join(". ", errors));
     }
+    Map<ModuleDescriptor, List<String>> permErrors = checkPermissionNames(modsAvailable,
+        modsEnabled);
+    for (Map.Entry<ModuleDescriptor,List<String>> ent : permErrors.entrySet()) {
+      for (String msg : ent.getValue()) {
+        logger.info("{}: {}", ent.getKey().getId(), msg);
+      }
+    }
```
julianladisch, line 668: Could be extracted into a method logPermissionErrors(modsAvailable, modsEnabled).

Could be wrapped into `if (logger.isInfoEnabled() { ... }`.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -788,4 +796,132 @@ static void sortTenantModules(List<TenantModuleDescriptor> tml,
...
+            if (unknownPermissions != null) {
+              unknownPermissions.add(subPermission);
+            }
+          }
+        }
+      }
+    }
+  }
+
+  private static void getDefinedPermissions(
```
julianladisch, line 917: `addPermissions` might be a better name because it returns void and doesn't get from `defined` but writes into it.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -788,4 +796,132 @@ static void sortTenantModules(List<TenantModuleDescriptor> tml,
...
+        for (ModuleDescriptor md : ent.getValue()) {
+          errors.computeIfAbsent(md, x -> new ArrayList<>()).add("Permission '"
+              + permissionName + "' defined in multiple modules: " + names);
+        }
+      }
+    }
+    return errors;
+  }
+
+  private static void getReferPermissions(
```
julianladisch, line 871: `addPermissions` might be a better name because it returns void and doesn't get from `refer*` but writes into them.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: I think the info should be warn and seems overkil to check for that being enabled. This code, is only executed rarely (compared to proxy calls).

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: ok

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: ok

julianladisch (approved):
Looks good, thanks!

adamdickmeiss:
> Looks good, thanks!

Thx for checking

PR #1196, opened 2022-02-10
Log info of module activation OKAPI-1072

adamdickmeiss:
Log elapsed time when calling tenant operation.

https://issues.folio.org/browse/OKAPI-1072


okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -385,6 +385,8 @@ private void waitTenantInit(Tenant tenant, ModuleInstance getInstance,
     if (!options.checkInvoke(md.getId())) {
       return Future.succeededFuture();
     }
+    long startTime = System.nanoTime();
```
julianladisch, line 388: System.currentTimeMillis() is good enough: https://stackoverflow.com/questions/351565

julianladisch (approved):
Otherwise looks good, thanks!

PR #1197, opened 2022-02-11
Fix timing calculation OKAPI-1072

adamdickmeiss:
Follow up.. Which fixes mix of nano and milliseconds !!


julianladisch (approved):
Looks good, thanks!

steveellis (approved):

PR #1198, opened 2022-02-11
Parallelise migration OKAPI-939

adamdickmeiss:
https://issues.folio.org/browse/OKAPI-939

Migration is parallel for all modules with dependencies satisfied (not just those with no requires).

Sorting always for Disable and enable in DepResolution.sortTenantModules (so that simulate shows the order).
But in reality that's not required as it's sorted anyway in jobRunPending later.

The sorting is completely reworked again (sortTenantModules). But at least the strategy is same for simulate phase and "real" migration phase.

I wished I could turn sortTenantModules and JobRunPending into one method as they are similar.


adamdickmeiss:
> > Should this be configurable - whether to perform migration in parallel?
> 
> Yes. Otherwise some installations might get overloaded causing timeouts. We don't know if there are modules with migrations that fail when run in parallel (for example assuming exclusive database access). And the dependency declarations might be incorrect.
> 
> A switch like parallelMigration=true is sufficient. A more advanced solution might be a configuration option parallelMigrations=3 that limits the number. parallelMigrations=1 is sequential.

Such a switch is now implemented with a default value of 1 for now.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -1050,6 +1035,99 @@ private ModuleDescriptor findSystemInterface(Tenant tenant, String interfaceName
...
 
+  private boolean isExclusive(ModuleDescriptor md) {
+    return md.getSystemInterface("_tenantPermissions") != null
+        || md.getAuthRoutingEntry() != null;
+  }
+
+  private boolean depsOK(TenantModuleDescriptor tm, ModuleDescriptor md,
+      Collection<ModuleDescriptor> modules, Set<String> allProvided) {
+
+    if (tm.getAction().equals(Action.disable)) {
```
julianladisch, line 1046: Can you replace `.equals` by `==` when comparing enums for readability, better type check, and avoiding NPE?

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -1050,6 +1035,99 @@ private ModuleDescriptor findSystemInterface(Tenant tenant, String interfaceName
...
+      List<TenantModuleDescriptor> tml, Map<String, ModuleDescriptor> modsAvailable,
+      Set<String> allProvided, InstallJob job, AtomicInteger running, AtomicBoolean exclusive,
+      Promise<Void> promise) {
+
+    Iterator<TenantModuleDescriptor> iterator = tml.iterator();
+    while (iterator.hasNext() && running.get() < options.getMaxParallel() && !exclusive.get()) {
+      TenantModuleDescriptor tm = iterator.next();
+      ModuleDescriptor md = modsAvailable.get(tm.getId());
+      if ((tm.getStage().equals(TenantModuleDescriptor.Stage.pending)
+          || tm.getStage().equals(TenantModuleDescriptor.Stage.deploy))
```
julianladisch, line 1062: == replacing .equals, consider importing TenantModuleDescriptor.Stage.

okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -1050,6 +1035,99 @@ private ModuleDescriptor findSystemInterface(Tenant tenant, String interfaceName
...
+          && depsOK(tm, md, getEnabledModules(t), allProvided)
+          && (!isExclusive(md) || running.get() == 0)) {
```
julianladisch, line 1064: This complex if with 5 expressions combined with `|| && && && ||` can be simplified by using guard clauses; this also reduces indentation of the body of the if:

```
      if (   tm.getStage() != TenantModuleDescriptor.Stage.pending
          && tm.getStage() != TenantModuleDescriptor.Stage.deploy ) {
        continue;
      }
      if (! depsOK(tm, md, getEnabledModules(t), allProvided)) {
        continue;
      }
      ...
```

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -735,7 +665,70 @@ static void installMaxIterations(
...
+                .collect(Collectors.joining(", ")));
+        if (DepResolution.moduleDepProvided(modules, allProvided, md)) {
+          logger.info("yes: adding {}", md.getId());
+          iterator.remove();
+          iterator = tml.iterator();
+          result.add(tm);
+          modules.add(md);
+          String moduleFrom = tm.getFrom();
+          if (moduleFrom != null) {
+            logger.info("yes: removing from {}", moduleFrom);
```
julianladisch, line 721: sortTenantModules should log on debug level only.

julianladisch (approved):
Looks good, thanks!

adamdickmeiss:
> Looks good, thanks!

Oh really? I'll look at the comments and address them tomorrow. Thx.

okapi-core/src/main/java/org/folio/okapi/bean/ModuleDescriptor.java
```diff
@@ -232,6 +233,23 @@ public void setOptional(InterfaceDescriptor[] optional) {
     return all;
   }
 
+  /**
+   * Get auth routing entry if such exist for module.
```
steveellis, line 237: Typo: change "exist" to "exists". Same below.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -128,7 +128,6 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
 
   /**
    * Check if md's required and optional interfaces are provided by set of modules.
-   * The interface version is not checked, only the interface name.
    * Only interfaces listed in allProvided are checked, this allows to exclude some optional
```
steveellis, line 132: "exclusion of" is a little more idiomatic

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -137,54 +136,56 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
...
+        return false;
       }
     }
     return true;
   }
 
-  static void topoSort(List<ModuleDescriptor> modules) {
-    List<ModuleDescriptor> result = new LinkedList<>();
+  /**
+   * Check if md's provided interfaces are required by any of modules.
```
steveellis, line 164: "any of the modules"

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -137,54 +136,56 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
...
+   * Only interfaces listed in allProvided are checked, this allows to exclude some optional
```
steveellis, line 165: "for the exclusion of"

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -137,54 +136,56 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
...
+   * interfaces from the check.
+   *
+   * @param modules     the modules that check against
```
steveellis, line 168: "The modules to check against." -- just add capital and period.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -137,54 +136,56 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
...
+   * @param md          module to check.
```
steveellis, line 169: "Module to check."

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -137,54 +136,56 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
...
+   * @return            true if modules do not use any of the provided interfaces; false otherwise
```
steveellis, line 170: Probably capitalize sentence and add period.

okapi-core/src/main/raml/okapi.raml
```diff
@@ -46,6 +46,66 @@ traits:
...
+          all problem(s) with module upgrade(s).
+        type: boolean
+        required: false
+        default: false
+      invoke:
+        description: |
+         Whether to invoke for tenant init/permissions/purge. Use
+          "true" to invoke, "false" to not invoke. Any other value
+          is a regular expression that is matched against the module ID.
+          If that pattern matches, invoke is performed (same as "true)".
```
steveellis, line 78: "true")

steveellis (commented):
Just noticed a few typos.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: I don't use capitals anywhere for parameters as first letter. So doing that now would be a mix of styles.

adamdickmeiss: Same story with ending period. The last sentence do not end in period.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: I'll remove the ending period instead.

PR #1200, opened 2022-02-18
Log in milliseconds follow-up OKAPI-1072

adamdickmeiss:
The original value in seconds is often 0 and it seems better
to log a more precise value of the time it takes.


julianladisch (approved):
Looks good, thanks!

PR #1201, opened 2022-02-18
Log launch descriptor upon deployment

adamdickmeiss:
Good to double-check environment passed to processes.


ArgonV (approved):
I think this will do the trick!

PR #1203, opened 2022-02-21
Fix install fails migrating from Juniper to Kiwi OKAPI-1075

adamdickmeiss:
Change strategy on conflicts and upgrade if possible on
same product and then turn to disable.

The problem in this case is ui-inventory-es that is left
over from Juniper and no longer part of Kiwi, but not really
replaced or disabled. The new behavior is to disable modules
with conflicts (that are not explicitly installed).


julianladisch (approved):
Looks good, thanks!

PR #1210, opened 2022-02-23
Parallel migration of modules OKAPI-939

adamdickmeiss:
2nd round of change with fixes for optional circulat dependencies.
A real case is mod-source-record-manager + mod-source-record-storage.
mod-source-record-manager uses many interfaces from storage, but
storage optionally depend on optional dependency from manager.

The fix actually reduces code size, because we don't care about
optional dependencies in activation.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
@@ -132,18 +132,13 @@ public final class DepResolution {
    * some optional interfaces from the check.
    *
    * @param modules     the modules to check against
-   * @param allProvided interfaces for all modules
    * @param md          module to check.
    * @return            true if interface requirements are met
    */
   public static boolean moduleDepProvided(
-      Collection<ModuleDescriptor> modules, Set<String> allProvided,
-      ModuleDescriptor md) {
+      Collection<ModuleDescriptor> modules, ModuleDescriptor md) {
 
-    for (InterfaceDescriptor req : md.getRequiresOptionalList()) {
-      if (!allProvided.contains(req.getId())) {
-        continue;
-      }
+    for (InterfaceDescriptor req : md.getRequiresList()) {
       boolean found = false;
       for (ModuleDescriptor md1 : modules) {
         InterfaceDescriptor[] providesList = md1.getProvidesList();
@@ -666,28 +661,12 @@ public final class DepResolution {
     if (!errors.isEmpty()) {
       throw new OkapiError(ErrorType.USER, String.join(". ", errors));
     }
-    // in reality this is not required as install sort as well
-    sortTenantModules(tml, modsAvailable, enabledModules, modsEnabled);
+    sortTenantModules(tml, modsAvailable, enabledModules);
+
   }
 
   static void sortTenantModules(List<TenantModuleDescriptor> tml,
-      Map<String, ModuleDescriptor> modsAvailable, Collection<ModuleDescriptor> modules,
-      Map<String, ModuleDescriptor> modsEnabled) {
-
-    Set<String> allProvided = new HashSet<>();
-    for (ModuleDescriptor md : modsEnabled.values()) {
-      for (InterfaceDescriptor descriptor : md.getProvidesList()) {
-        if (descriptor.isRegularHandler()) {
-          allProvided.add(descriptor.getId());
-        }
-      }
-    }
-    sortTenantModules(tml, modsAvailable, modules, allProvided);
-  }
-
-  static void sortTenantModules(List<TenantModuleDescriptor> tml,
-      Map<String, ModuleDescriptor> modsAvailable, Collection<ModuleDescriptor> modules,
-      Set<String> allProvided) {
+      Map<String, ModuleDescriptor> modsAvailable, Collection<ModuleDescriptor> modules) {
 
     logger.info("sortTenantModules with list {}",
         tml.stream().map(TenantModuleDescriptor::getId).collect(Collectors.joining(", ")));
@@ -711,7 +690,7 @@ public final class DepResolution {
         logger.debug("See if module {} can be added to existing list of modules {}",
             md.getId(), modules.stream().map(ModuleDescriptor::getId)
                 .collect(Collectors.joining(", ")));
-        if (DepResolution.moduleDepProvided(modules, allProvided, md)) {
+        if (DepResolution.moduleDepProvided(modules, md)) {
           logger.debug("yes: adding {}", md.getId());
           iterator.remove();
           iterator = tml.iterator();
@@ -728,6 +707,10 @@ public final class DepResolution {
         iterator = tml.iterator();
         result.add(tm);
       }
+    }
+    if (!tml.isEmpty()) {
+      throw new OkapiError(ErrorType.USER, "Some modules cannot be topological sorted: "
+          + tml.stream().map(TenantModuleDescriptor::getId).collect(Collectors.joining(", ")));
     }
     tml.addAll(result);
   }
--- a/okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
@@ -127,38 +127,32 @@ public final class DepResolution {
   }
 
   /**
-   * Check if md's required and optional interfaces are provided by set of modules.
-   * Only interfaces listed in allProvided are checked, this allows for the exclusion of
-   * some optional interfaces from the check.
+   * Check if md's required interfaces are provided by set of modules.
    *
    * @param modules     the modules to check against
    * @param md          module to check.
-   * @return            true if interface requirements are met
+   * @return            true if interface requirements are met; false otherwise
    */
   public static boolean moduleDepProvided(
       Collection<ModuleDescriptor> modules, ModuleDescriptor md) {
 
+    interfaceDescriptor:
     for (InterfaceDescriptor req : md.getRequiresList()) {
-      boolean found = false;
       for (ModuleDescriptor md1 : modules) {
         InterfaceDescriptor[] providesList = md1.getProvidesList();
         for (InterfaceDescriptor prov : providesList) {
           if (prov.isRegularHandler() && prov.isCompatible(req)) {
-            found = true;
+            continue interfaceDescriptor;
           }
         }
       }
-      if (!found) {
-        return false;
-      }
+      return false;
     }
     return true;
   }
 
   /**
-   * Check if md's provided interfaces are required by any of the modules.
-   * Only interfaces listed in allProvided are checked, this allows for the exclusion of
-   * some optional interfaces from the check.
+   * Check if at least one of md's provided interfaces is a required or optional dependency of any of the modules.
    *
    * @param modules     the modules to check against
    * @param md          module to check
@@ -665,10 +659,20 @@ public final class DepResolution {
 
   }
 
+  /**
+   * Sort the modules honoring enable/disable actions.
+   * <p>This is topological sort where nodes represent modules and arcs represent
+   * interface dependencies.
+   * </p>
+   * @param tml the module list with actions and the resulting sorted list afterwards.
+   * @param modsAvailable all known modules
+   * @param modules the existing list of modules and current list as we go on
+   * @throws OkapiError if dependencies can not be satisfied - including circular dependencies.
+   */
   static void sortTenantModules(List<TenantModuleDescriptor> tml,
       Map<String, ModuleDescriptor> modsAvailable, Collection<ModuleDescriptor> modules) {
 
-    logger.info("sortTenantModules with list {}",
+    logger.info("sortTenantModules with list {}", () ->
         tml.stream().map(TenantModuleDescriptor::getId).collect(Collectors.joining(", ")));
     List<TenantModuleDescriptor> result = new ArrayList<>();
     Iterator<TenantModuleDescriptor> iterator = tml.iterator();
@@ -709,6 +713,7 @@ public final class DepResolution {
       }
     }
     if (!tml.isEmpty()) {
+      // it would be good to analyze this further with the interfaces that are problematic
       throw new OkapiError(ErrorType.USER, "Some modules cannot be topological sorted: "
           + tml.stream().map(TenantModuleDescriptor::getId).collect(Collectors.joining(", ")));
     }
@@ -761,7 +766,7 @@ public final class DepResolution {
    * Find modules that provide a required interface.
    * @param modsAvailable all modules known
    * @param req required interface to check
-   * @return newest modules that meets the requirements; empty if no modules are found
+   * @return newest modules that meet the requirements; empty if no modules are found
    */
   static Map<String, ModuleDescriptor> findModulesForRequiredInterface(
       Map<String, ModuleDescriptor> modsAvailable, InterfaceDescriptor req) {
--- a/okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
@@ -152,7 +152,8 @@ public final class DepResolution {
   }
 
   /**
-   * Check if at least one of md's provided interfaces is a required or optional dependency of any of the modules.
+   * Check if any of md's provided interfaces are used by any of the modules.
+   * <p>A interface is used if it's a required or optional interface dependency.</p>
    *
    * @param modules     the modules to check against
    * @param md          module to check
--- a/okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
+++ b/okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
@@ -665,10 +665,11 @@ public final class DepResolution {
    * <p>This is topological sort where nodes represent modules and arcs represent
    * interface dependencies.
    * </p>
+   * @see <a href="https://en.wikipedia.org/wiki/Topological_sorting">Topological sorting</a>
    * @param tml the module list with actions and the resulting sorted list afterwards.
    * @param modsAvailable all known modules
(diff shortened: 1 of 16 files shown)


okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -128,63 +128,59 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
...
-            if (prov.isRegularHandler() && prov.getId().equals(req.getId())) {
-              found = true;
-            }
+    for (InterfaceDescriptor req : md.getRequiresList()) {
+      boolean found = false;
+      for (ModuleDescriptor md1 : modules) {
+        InterfaceDescriptor[] providesList = md1.getProvidesList();
+        for (InterfaceDescriptor prov : providesList) {
+          if (prov.isRegularHandler() && prov.isCompatible(req)) {
+            found = true;
```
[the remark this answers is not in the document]
steveellis: Or perhaps I'm misunderstanding what that outer for loop is doing. In any case the logic is a bit hard to follow.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -128,63 +128,59 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
 
   /**
    * Check if md's required and optional interfaces are provided by set of modules.
-   * The interface version is not checked, only the interface name.
-   * Only interfaces listed in allProvided are checked, this allows to exclude some optional
-   * interfaces from the check.
+   * Only interfaces listed in allProvided are checked, this allows for the exclusion of
```
julianladisch, line 131: There's no `allProvided` any more.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -736,7 +661,58 @@ static void installMaxIterations(
     if (!errors.isEmpty()) {
       throw new OkapiError(ErrorType.USER, String.join(". ", errors));
     }
-    sortTenantModules(tml, modsAvailable, modsEnabled);
+    sortTenantModules(tml, modsAvailable, enabledModules);
+
+  }
+
+  static void sortTenantModules(List<TenantModuleDescriptor> tml,
```
steveellis, line 668: Might be good to have a javadoc here. I'm left wondering what the sort is for. The error message below says its a "topological sort" -- seems important what that is. If it is topological what's the topic, etc.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -128,63 +128,59 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
 
   /**
    * Check if md's required and optional interfaces are provided by set of modules.
```
julianladisch, line 130: This implementation doesn't check optional interfaces.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -784,8 +760,8 @@ static void installMaxIterations(
   /**
    * Find modules that provide a required interface.
    * @param modsAvailable all modules known
-   * @param req required interface to check.
-   * @return newest modules that meets the requirements; empty if no modules are found.
+   * @param req required interface to check
+   * @return newest modules that meets the requirements; empty if no modules are found
```
steveellis, line 764: "modules that that meet the requirements"

okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
```diff
@@ -103,4 +106,19 @@ public boolean getReinstall() {
   public void setReinstall(boolean reinstall) {
     this.reinstall = reinstall;
```
steveellis, line 107: `this` can be removed and other setters in the class don't use it.

okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
```diff
@@ -103,4 +106,19 @@ public boolean getReinstall() {
...
+
+  /**
+   * Set max parallel tenant invocations.
+   * @param maxParallel positive integer
+   */
+  public void setMaxParallel(int maxParallel) {
+    if (maxParallel < 1) {
+      throw new OkapiError(ErrorType.USER, "parallel must be 1 or higher");
+    }
+    this.maxParallel = maxParallel;
```
steveellis, line 122: Same as above

okapi-core/src/main/raml/okapi.raml
```diff
@@ -46,6 +46,66 @@ traits:
...
+      deploy:
+        description: Whether to deploy (or undeploy if disabling)
+        type: boolean
+        required: false
+        default: false
+      ignoreErrors:
+        description: |
+          Okapi 4.2.0 and later, it is possible to ignore errors during the
+          install operation. This is done by supplying parameter `ignoreErrors=true`.
+          In this case, Okapi will try to upgrade all
```
steveellis, line 65: Line breaks could be made more consistent, maybe at 100 chars.

okapi-core/src/main/raml/okapi.raml
```diff
@@ -46,6 +46,66 @@ traits:
...
+        type: boolean
+        required: false
+        default: false
+      simulate:
+        description: Whether the installation is simulated
+        type: boolean
+        required: false
+        default: false
+      tenantParameters:
+        description: Parameters for Tenant init
```
steveellis, line 106: "Tenant" probably doesn't need to be capitalized. It isn't in other properties here.

okapi-core/src/main/raml/okapi.raml
```diff
@@ -962,7 +1022,8 @@ traits:
               text/plain:
       post:
         is: [
-          versionFilters
+          versionFilters,
+          installParameters
         ]
         description: Enable, disable or upgrade one or more modules for
           tenant. The request body and response body is of the same type
```
steveellis, line 1029: Probably need a period at the end of the sentence here.

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
```diff
@@ -2010,14 +2010,14 @@ public void testRequestOnly(TestContext context) {
...
         .body(equalTo("[ {" + LS
-            + "  \"id\" : \"request-only-1.0.0\"," + LS
-            + "  \"action\" : \"disable\"" + LS
-            + "}, {" + LS
             + "  \"id\" : \"request-pre-1.0.0\"," + LS
             + "  \"action\" : \"enable\"" + LS
             + "}, {" + LS
             + "  \"id\" : \"auth-f-module-1\"," + LS
             + "  \"action\" : \"enable\"" + LS
+            + "}, {" + LS
```
steveellis, line 2018: Matching like this seems a bit prone to break since it relies on whitespace, line breaks and such being just right (at least if I'm understanding how `equalTo` treats it). Wouldn't it be better to parse the object and check its properties?

But please ignore. I can see this kind of matching is used extensively and I can see how addresses the problem of "I want this complex object with these properties". Anyway just curious about the history of this.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -128,63 +128,59 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
...
+        return false;
       }
     }
     return true;
   }
 
-  static void topoSort(List<ModuleDescriptor> modules) {
-    List<ModuleDescriptor> result = new LinkedList<>();
+  /**
+   * Check if md's provided interfaces are required by any of the modules.
```
julianladisch, line 159: Check if at least one of md's provided interfaces is a required or optional dependency of any of the modules.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -736,7 +661,58 @@ static void installMaxIterations(
...
     }
-    sortTenantModules(tml, modsAvailable, modsEnabled);
+    sortTenantModules(tml, modsAvailable, enabledModules);
+
+  }
+
+  static void sortTenantModules(List<TenantModuleDescriptor> tml,
+      Map<String, ModuleDescriptor> modsAvailable, Collection<ModuleDescriptor> modules) {
+
+    logger.info("sortTenantModules with list {}",
```
julianladisch, line 671:
```suggestion
    logger.info("sortTenantModules with list {}", () ->
```

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -736,7 +661,58 @@ static void installMaxIterations(
...
+          }
+        }
+      } else {
+        iterator.remove();
+        iterator = tml.iterator();
+        result.add(tm);
+      }
+    }
+    if (!tml.isEmpty()) {
+      throw new OkapiError(ErrorType.USER, "Some modules cannot be topological sorted: "
```
julianladisch, line 712: These modules cannot be topological sorted because they have circular or unsatisfiable interface dependencies:

adamdickmeiss: Exactly!

okapi-core/src/test/java/org/folio/okapi/InstallTest.java
```diff
@@ -1514,4 +1519,353 @@ public void installTenantInitVersion2JobError(TestContext context) {
...
+        .header("Content-Type", "application/json")
+        .body(mid.encode()).post("/_/proxy/modules").then().statusCode(201);
+
+    final String tenant = "roskilde";
+    createTenant(context, tenant);
+    deployAsyncInitModule(context, moda.getString("id"), portModule);
+    deployAsyncInitModule(context, modb.getString("id"), portModule);
+    deployAsyncInitModule(context, mid.getString("id"), portModule);
+    ModuleTenantInitAsync tModule = new ModuleTenantInitAsync(vertx, portModule);
+    tModule.start().onComplete(context.asyncAssertSuccess());
```
julianladisch, line 1564:
```
    Async start = context.async();
    tModule.start().onComplete(context.asyncAssertSuccess(x -> start.complete()));
    start.await();
```
Otherwise the following post might run before start has completed.

okapi-core/src/test/java/org/folio/okapi/InstallTest.java
```diff
@@ -1514,4 +1519,353 @@ public void installTenantInitVersion2JobError(TestContext context) {
...
+            .add(new JsonObject()
+                .put("id", mid.getString("id"))
+                .put("action", "enable")
+                .put("stage", "done")
+            )
+        );
+    context.assertEquals(jobExpected, job);
+    // moda , modb running in parallel
+    context.assertTrue(tModule.getStartTime(moda.getString("id")).isBefore(tModule.getEndTime(modb.getString("id"))));
+    context.assertTrue(tModule.getStartTime(modb.getString("id")).isBefore(tModule.getEndTime(moda.getString("id"))));
```
julianladisch, line 1621: `assertParallelRun(ModuleTenantInitAsync a, ModuleTenantInitAsync b)` and
`assertSequentialRun(ModuleTenantInitAsync a, ModuleTenantInitAsync b)`
can avoid duplicate code.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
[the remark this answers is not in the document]
adamdickmeiss: Returning true here would only check for one interface (not all).

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: yep and optional not checked. Fixed.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -128,63 +128,59 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
...
-        if (!found) {
-          return false;
-        }
+      }
+      if (!found) {
+        return false;
       }
     }
     return true;
   }
```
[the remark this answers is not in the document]
adamdickmeiss: neat

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: Agreed. Added some javadoc to the method. This is fun stuff :-)

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: Correct. Fixed now!

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: Oh yes.

okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
adamdickmeiss: I don't understand what you mean.

okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
adamdickmeiss: Also here I don't understand.

okapi-core/src/main/raml/okapi.raml
adamdickmeiss: Yeah.. It's all over the place.

okapi-core/src/main/raml/okapi.raml
adamdickmeiss: sure

okapi-core/src/main/raml/okapi.raml
adamdickmeiss: It's there on next line.. But I can move it up.

okapi-core/src/test/java/org/folio/okapi/ProxyTest.java
adamdickmeiss: Yes.. Should use JsonObject .. in a lot of places.. Another PR with just that one day.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: "of any of the" triggers a warning here :-) I understand it.

okapi-core/src/test/java/org/folio/okapi/InstallTest.java
adamdickmeiss: sure

okapi-core/src/test/java/org/folio/okapi/InstallTest.java
adamdickmeiss: ok

okapi-core/src/main/raml/okapi.raml
```diff
@@ -46,6 +46,66 @@ traits:
         enum: ["true","false","only"]
         required: false
         default: "true"
+  installParameters:
+    queryParameters:
+      async:
+        description: Whether to install in the background
```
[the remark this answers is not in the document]
adamdickmeiss: I'll make a separate PR on the RAML

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
```diff
@@ -127,64 +127,55 @@ public static void getLatestProducts(int limit, List<ModuleDescriptor> mdl) {
...
+      return false;
     }
     return true;
   }
 
-  static void topoSort(List<ModuleDescriptor> modules) {
-    List<ModuleDescriptor> result = new LinkedList<>();
+  /**
+   * Check if any of md's provided interfaces are used by any of the modules.
+   * <p>A interface is used if it's a required or optional interface dependency.</p>
```
julianladisch, line 156: The closing p HTML tag is not required because javadoc uses HTML, not XHTML.
It is very uncommon to use the closing p HTML tag. Therefore some style guides explicitly disallow `</p>`:
https://blog.joda.org/2012/11/javadoc-coding-standards.html
https://metaterasology.github.io/docs/workflow/documentationJavaDoc.html#html-tags
Using `</p>` is uncommon in FOLIO javadoc and Okapi javadoc.

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
julianladisch: Can you add the full explanation to the error message?

julianladisch (approved):
Looks good, thanks!

okapi-core/src/main/java/org/folio/okapi/util/DepResolution.java
adamdickmeiss: ok.

okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
steveellis: I didn't notice that the param name was the same as the variable name, so you do need `this`.

okapi-core/src/main/java/org/folio/okapi/util/TenantInstallOptions.java
steveellis: Same

steveellis (approved):

PR #1213, opened 2022-02-27
Enable HTTP compression for pull OKAPI-1079

adamdickmeiss:


adamdickmeiss:
Closing as it's not it's not a good idea to make Okapi perform compression of regular responses to modules. It's better as it's been until now to avoid Okapi inspecting buffers that is passed to/from modules.

PR #1214, opened 2022-02-28
Compress Okapi responses > 16K OKAPI-1079

adamdickmeiss:
HttpClient for pull operation has compression enabled.

With a client in Europe and a server in US, I see a speed again of about 30% for the pull operation as a whole with this.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -990,7 +990,7 @@ public class ProxyService {
     clientsEnd(bcontent, clientRequestList);
     try {
       internalModule.internalService(req, pc)
-          .compose(resp -> {
+          .map(resp -> {
             int statusCode = pc.getCtx().response().getStatusCode();
             if (statusCode == 200 && resp.isEmpty()) {
               // Say "no content", if there isn't any
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -961,16 +961,16 @@ public class ProxyService {
 
   static Buffer compressInternalResponse(RoutingContext ctx, Buffer respBuf) {
     String acceptEncoding = ctx.request().getHeader("Accept-Encoding");
-    int sz = respBuf.length();
+    int uncompressedSize = respBuf.length();
     if (acceptEncoding != null && acceptEncoding.contains("gzip")
-        && sz > CONTENT_COMPRESS_MIN_SIZE) {
+        && uncompressedSize > CONTENT_COMPRESS_MIN_SIZE) {
       try {
-        ByteArrayOutputStream dst = new ByteArrayOutputStream();
-        GZIPOutputStream src = new GZIPOutputStream(dst);
-        src.write(respBuf.getBytes());
-        src.close();
-        respBuf = Buffer.buffer(dst.toByteArray());
-        logger.info("Compressed {} {}", sz, respBuf.length());
+        ByteArrayOutputStream compressedStream = new ByteArrayOutputStream();
+        GZIPOutputStream uncompressedStream = new GZIPOutputStream(compressedStream);
+        uncompressedStream.write(respBuf.getBytes());
+        uncompressedStream.close();
+        respBuf = Buffer.buffer(compressedStream.toByteArray());
+        logger.info("Compressed {} bytes to {} bytes", uncompressedSize, respBuf.length());
         ctx.response().putHeader("Content-Encoding", "gzip");
       } catch (IOException e) {
         throw new UncheckedIOException(e);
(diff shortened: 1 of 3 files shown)


okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -954,19 +959,38 @@ private void proxyInternal(Iterator<ModuleInstance> it,
     );
   }
 
+  static Buffer compressInternalResponse(RoutingContext ctx, Buffer respBuf) {
+    String acceptEncoding = ctx.request().getHeader("Accept-Encoding");
+    int sz = respBuf.length();
```
steveellis, line 964: Minor: maybe the variable name here `sz` could be more clear like `bufferLength` or just `size`.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -954,19 +959,38 @@ private void proxyInternal(Iterator<ModuleInstance> it,
...
+    if (acceptEncoding != null && acceptEncoding.contains("gzip")
+        && sz > CONTENT_COMPRESS_MIN_SIZE) {
+      try {
+        ByteArrayOutputStream dst = new ByteArrayOutputStream();
```
steveellis, line 968: Seems like `dst` should be some other name like `outputStream` or similar.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -954,19 +959,38 @@ private void proxyInternal(Iterator<ModuleInstance> it,
...
+        GZIPOutputStream src = new GZIPOutputStream(dst);
```
steveellis, line 969: Sorry I'm on a roll now... :) Maybe `src` should be something like `gzipedOutputStream` or `compressedOutput` or similar.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: I dunno.. both `src` and `dst` are outputstreams. `uncompressedOutputStream` and `compressedOutputStream`??

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: ok

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: ok. I've renamed them now

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -954,19 +959,38 @@ private void proxyInternal(Iterator<ModuleInstance> it,
...
+        src.write(respBuf.getBytes());
+        src.close();
+        respBuf = Buffer.buffer(dst.toByteArray());
+        logger.info("Compressed {} {}", sz, respBuf.length());
```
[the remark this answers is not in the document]
adamdickmeiss: ok

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -954,19 +959,38 @@ private void proxyInternal(Iterator<ModuleInstance> it,
     );
   }
 
+  static Buffer compressInternalResponse(RoutingContext ctx, Buffer respBuf) {
```
julianladisch, line 962: What advantages has this code over Vert.x'/Netty's [setCompressionSupported](https://vertx.io/docs/apidocs/io/vertx/core/http/HttpServerOptions.html#setCompressionSupported-boolean-)?
https://vertx.io/docs/vertx-core/java/#_http_compression

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
steveellis: Seems like `ByteArrayOutputStream` has a `close` method. I haven't looked at its source but just googling on it there seems to be a difference of opinion on whether to call it does anything since the implementation may not do anything. FWIW someone seems to think it's still a good idea to call in case the implementation changes in future releases. Also FWIW unclosed streams are bad in my experience. Could be the same for the stream that's doing the zipping.

okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
adamdickmeiss: Only that okapi does not try to compress responses from modules.. In this PR that choice is left to the module to decide. Just as Okapi doesn't do for own handled responses less than some size. I'll make a branch with server compresion as well (did that in #1213 but that *also* used http compression for Okapi client).. The worst for all parties is if okapi uncompresses responses from backend modules.. then compressed them again.. then let the client calling okapi decompress .. More CPU work for no gain.

adamdickmeiss:
Closing .. doing #1216 instead

PR #1215, opened 2022-03-02
Unit test fixes for slow server

adamdickmeiss:
This PR tunes tests a bit so Okapi can complete its test with a relatively slow server.

The slow server being based on AMD A10 Micro-6700T.


steveellis (approved):
Cool use of Awaitility and interesting tests. But seems fine if all we're doing is waiting longer to compensate for slow machines.

PR #1217, opened 2022-03-03
Reject invalid tenant ids OKAPI-1081

adamdickmeiss:
https://issues.folio.org/browse/OKAPI-1081


julianladisch (changes requested):
This cannot been code reviewed because we haven't agreed on the requirements.

The requirements need to be discussed with the implementers and sysops before.

julianladisch (approved):
Looks good, thanks!

PR #1219, opened 2022-03-04
Restrict module name OKAPI-1082

adamdickmeiss:
Restrict module name (AKA product) to begin with letter and only
include alphanumeric characters as well as underscore and hyphen-minus.


marcjohnson-kint:
> Restrict module name (AKA product) to begin with letter and only include alphanumeric characters as well as underscore and hyphen-minus.

This is a breaking change (in the sense that something that was valid before no longer is) will it be included in a new major version of Okapi?

adamdickmeiss:
> > Restrict module name (AKA product) to begin with letter and only include alphanumeric characters as well as underscore and hyphen-minus.
> 
> This is a breaking change (in the sense that something that was valid before no longer is) will it be included in a new major version of Okapi?

This will be include in next release (but not in 4.13 of course). And next version should be called 5.0.0.

marcjohnson-kint:
> This will be include in next release (but not in 4.13 of course). And next version should be called 5.0.0.

Ok. As it stands the next version is 4.14.0 (not 5.0.0).

adamdickmeiss:
> > This will be include in next release (but not in 4.13 of course). And next version should be called 5.0.0.
> 
> Ok. As it stands the next version is 4.14.0 (not 5.0.0).



> > This will be include in next release (but not in 4.13 of course). And next version should be called 5.0.0.
> 
> Ok. As it stands the next version is 4.14.0 (not 5.0.0).

Yes. I just want to see if there are other views on this , before proceeding.

julianladisch (changes requested):
This cannot been code reviewed because we haven't agreed on the requirements.

The requirements need to be discussed with the implementers and sysops.

PR #1221, opened 2022-03-07
Remove ExtendedAsyncResult OKAPI-1078

adamdickmeiss:
OkapiClient returns AsyncResult rather than ExtendedAsyncResult.
Use ErrorTypeException.getType to check for ErrorType.


julianladisch (approved):
Looks good, thanks!

PR #1222, opened 2022-03-09
Property enable_system_auth checked for proxy calls

adamdickmeiss:
This property which should only be set to false for testing
purposes is now also checked when phase auth is called in proxy.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
+++ b/okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
@@ -230,7 +230,8 @@ public class ProxyService {
       logger.debug("getModulesForRequest: Checking {} '{}' '{}'",
           re.getPathPattern(), phase, re.getLevel());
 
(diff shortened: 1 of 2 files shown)


okapi-core/src/main/java/org/folio/okapi/managers/ProxyService.java
```diff
@@ -510,7 +511,7 @@ private void relayToRequest(HttpClientResponse res, ProxyContext pc,
   }
 
   private void log(HttpClientRequest creq) {
-    logger.debug("{} {}", creq.getMethod().name(), creq.getURI());
+    logger.debug("{} {}", () -> creq.getMethod().name(), () -> creq.getURI());
```
julianladisch, line 514: This is not needed if using
```
if (! logger.isDebugEnabled()) {
  return;
}
```
at the begin of this method.

adamdickmeiss: OK.. Lower coverage though ..

julianladisch (approved):
Looks good, thanks!

PR #1228, opened 2022-04-01
Move jackson-databind entry before jackson-bom

adamdickmeiss:
The first entry takes precendence over later entries in pom.xml.


julianladisch (approved):
Looks good, thanks!

PR #1231, opened 2022-04-11
Fix exception SemVer with 4000001006 OKAPI-1091

adamdickmeiss:
Switch to Long for SemVer component.

https://issues.folio.org/browse/OKAPI-1091


julianladisch (approved):
Looks good, thanks!

PR #1238, opened 2022-05-04
Enable SCRAM-SHA-256 PostgreSQL passwords OKAPI-793

adamdickmeiss:


okapi-core/pom.xml
```diff
@@ -84,6 +84,10 @@
       <groupId>com.cronutils</groupId>
       <artifactId>cron-utils</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.ongres.scram</groupId>
+      <artifactId>client</artifactId>
+    </dependency>
```
julianladisch, line 90: Can you add a test to https://github.com/folio-org/okapi/blob/master/okapi-core/src/test/java/org/folio/okapi/service/impl/PostgresHandleTest.java that shows that SCRAM-SHA-256 actually works?

adamdickmeiss: > Can you add a test to https://github.com/folio-org/okapi/blob/master/okapi-core/src/test/java/org/folio/okapi/service/impl/PostgresHandleTest.java that shows that SCRAM-SHA-256 actually works?

Hoped you never asked :-)

okapi-core/src/test/java/org/folio/okapi/service/impl/PostgresHandleTest.java
```diff
@@ -127,4 +127,17 @@ void rejectTlsv1_2(Vertx vertx, VertxTestContext vtc) {
...
+    configure("ssl = on", "password_encryption = scram-sha-256");
+    new PostgresHandle(vertx, config()).getConnection()
+        .compose(connection -> {
+          String sql = "ALTER USER " + POSTGRESQL_CONTAINER.getUsername()
+              + " WITH PASSWORD '" + POSTGRESQL_CONTAINER.getPassword() + "';";
+          return connection.query(sql).execute().mapEmpty();
+        })
+        .compose(x -> new PostgresHandle(vertx, config()).getConnection())
+        .onComplete(vtc.succeedingThenComplete());
+  }
```
julianladisch, line 144: `POSTGRESQL_CONTAINER.getUsername()` is used by other test methods and should not be touched.
The `pg_authid` table reveals the actual method.
```
  @Test
  void scram256(Vertx vertx, VertxTestContext vtc) {
    configure("password_encryption = scram-sha-256");
    JsonObject config = config();
    config.remove("postgres_server_pem");
    new PostgresHandle(vertx, config()).getConnection()
        .compose(connection -> {
            return connection.query("CREATE USER scram PASSWORD 'foo'").execute()
                .compose(x -> connection.query("SELECT rolpassword FROM pg_authid WHERE rolname='scram'").execute())
                .map(rowSet -> rowSet.iterator().next().getString(0));
        })
        .onComplete(vtc.succeeding(password -> assertThat(password).startsWith("SCRAM-SHA-256")))
        .compose(x -> {
          config.put("postgres_username", "scram").put("postgres_password", "foo");
          return new PostgresHandle(vertx, config).getConnection();
        })
        .onComplete(vtc.succeedingThenComplete());
```

adamdickmeiss: ok. Disabling ssl did not work .. and, frankly, why should we check this.

julianladisch: Disabling ssl works if replacing config() by config in line 138.

julianladisch (approved):
Looks good, thanks!

PR #1241, opened 2022-05-11
Fix javadoc for HttpClientFactory methods

adamdickmeiss:


julianladisch (approved):
Good catch, thank you!

PR #1250, opened 2022-07-12
Fix mod-authtoken tenant init not called OKAPI-1111

adamdickmeiss:


steveellis (approved):
Thank you Adam for taking this on.

PR #1251, opened 2022-07-15
Cookie handling OKAPI-1104

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/util/TokenHeader.java
```diff
@@ -0,0 +1,73 @@
...
+   * <p>Based on RFC6265.
+   * <a href="https://www.rfc-editor.org/rfc/rfc6265#section-4.2.1">Cookie syntax</a>
+   *    <a href="https://www.rfc-editor.org/rfc/rfc6265#section-4.1.1">Cookie pair syntax</a>
+   * @param cookie cookie string
+   * @param name name to look for
+   * @return value; null if not found
+   * @throws IllegalArgumentException if name occurs multiple times.
+   */
+  static String cookieValue(String cookie, String name) {
+    String[] components = cookie.trim().split("; ");
```
julianladisch, line 22: Using existing parser `ServerCookieDecoder.STRICT.decodeAll(cookie)` avoids extensive tests of edge cases.

adamdickmeiss: Ok. Will look into that

okapi-common/src/main/java/org/folio/okapi/common/XOkapiHeaders.java
```diff
@@ -197,4 +197,10 @@ private XOkapiHeaders() {
    * version.
    */
   public static final String OKAPI_MODULE = "okapi";
+
+  /**
+   * Cookie acess token name.
+   */
+  public static final String COOKIE_ACCESS_TOKEN = "accessToken";
```
julianladisch, line 204: wiki.example.com might set an accessToken for *.example.com that will be set for okapi.example.com.

To avoid a name clash the cookie name should be more specific, for example folioAccessToken.

steveellis: I made this change and pushed it here so that we could have all the branches aligned for testing. See what you think: https://github.com/folio-org/okapi/pull/1251/commits/ed437e12699d3bad1c6ba45ed2342e5f6f85f890.

Going to now make this change in the other PRs that are affected.

steveellis (approved):

PR #1256, opened 2022-08-25
Use log4j map lookup; fixes OKAPI-1117

adamdickmeiss:


okapi-core/src/main/resources/log4j2.properties
```diff
@@ -1,19 +1,25 @@
...
 filters = threshold
 
 filter.threshold.type = ThresholdFilter
 filter.threshold.level = info
 
-appenders = console
+appender.full.type = Console
+appender.full.name = FULL
+appender.full.layout.type = PatternLayout
+appender.full.layout.pattern = %d{HH:mm:ss} [${map:requestid}] [${map:tenantid}] [${map:userid}] [${map:moduleid}] %-5p %-20.20C{1} %m%n
```
steveellis, line 12: Maybe instead of `%d{HH:mm:ss}` we should do `%d{DEFAULT} ` which includes the date and microseconds like this `2012-11-02 14:34:02,123`. There are other possibilities: https://logging.apache.org/log4j/2.x/manual/layouts.html. I think in a lot of envs (like the rancher ones) this file is just taken as is and having the additional info could help debug.

It also might help self-hosters who also may just take this file without much thought and then be surprised that they can't see dates in their logs.

okapi-core/src/main/resources/log4j2.properties
```diff
@@ -1,19 +1,25 @@
...
 
-appender.console.type = Console
-appender.console.name = STDOUT
-appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = %d{HH:mm:ss} [$${FolioLoggingContext:requestid}] [$${FolioLoggingContext:tenantid}] [$${FolioLoggingContext:userid}] [$${FolioLoggingContext:moduleid}] %-5p %-20.20C{1} %m%n
+appender.omitted.type = Console
+appender.omitted.name = OMITTED
+appender.omitted.layout.type = PatternLayout
+appender.omitted.layout.pattern = %d{HH:mm:ss} [] [] [] [] %-5p %-20.20C{1} %m%n
```
steveellis, line 17: Same recommendation as above to include date for the omitted pattern.

okapi-core/src/main/resources/log4j2.properties
adamdickmeiss: I'm ok with that.. @skoczko ? @julianladisch ?

okapi-core/src/main/java/org/folio/okapi/util/ProxyContext.java
```diff
@@ -25,6 +26,8 @@
 public class ProxyContext {
 
   private static final Logger logger = OkapiLogger.get();
+
```
julianladisch, line 29: This empty line should be after fullLogger, not before.

okapi-core/src/main/java/org/folio/okapi/util/ProxyContext.java
```diff
@@ -25,6 +26,8 @@
 public class ProxyContext {
 
   private static final Logger logger = OkapiLogger.get();
```
julianladisch, line 28:
```suggestion
  private static final Logger logger = OkapiLogger.get();  // logger name "okapi"
```
This is different from log4j's default name: LoggerManager.getLogger() uses the caller class ProxyContext.

okapi-core/src/main/resources/log4j2.properties
julianladisch: The PatternLayout should be consistent across all FOLIO modules.
Currently we always use `%d{HH:mm:ss}`: https://github.com/search?l=Java+Properties&q=org%3Afolio-org+PatternLayout&type=Code
When we change this default it should be coordinated and changed in all modules, and planned for a specific flower release, similar to https://wiki.folio.org/display/TC/Officially+Supported+Technologies
Therefore the PatternLayout change should not be included in this PR.
Please create a FOLIO Jira and discuss it in the #sys-ops Slack channel.
Inplementers can always provide a log4j2.properties file that overwrites the FOLIO default.

okapi-core/src/main/resources/log4j2.properties
```diff
@@ -1,19 +1,25 @@
...
+appender.full.name = FULL
+appender.full.layout.type = PatternLayout
+appender.full.layout.pattern = %d{HH:mm:ss} [${map:requestid}] [${map:tenantid}] [${map:userid}] [${map:moduleid}] %-5p %-20.20C{1} %m%n
 
-appender.console.type = Console
-appender.console.name = STDOUT
-appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = %d{HH:mm:ss} [$${FolioLoggingContext:requestid}] [$${FolioLoggingContext:tenantid}] [$${FolioLoggingContext:userid}] [$${FolioLoggingContext:moduleid}] %-5p %-20.20C{1} %m%n
+appender.omitted.type = Console
+appender.omitted.name = OMITTED
```
julianladisch, line 15: For me `omitted` is somewhat misleading because the log entry is not omitted. What about `basic`?

okapi-core/src/main/java/org/folio/okapi/util/ProxyContext.java
```diff
@@ -203,20 +204,20 @@ public MultiMap getHandlerHeaders() {
...
-    StringBuilder mods = new StringBuilder();
-    if (modList != null && !modList.isEmpty()) {
-      for (ModuleInstance mi : modList) {
-        mods.append(" ").append(mi.getModuleDescriptor().getId());
-      }
+    String mods = "";
+    if (modList != null) {
+      mods = modList.stream().map(x -> x.getModuleDescriptor().getId())
+          .collect(Collectors.joining(" "));
     }
```
julianladisch, line 214: The mods calculation should be moved into the isInfoEnabled() if clause.

okapi-core/src/main/java/org/folio/okapi/util/ProxyContext.java
```diff
@@ -203,20 +204,20 @@ public MultiMap getHandlerHeaders() {
...
     if (logger.isInfoEnabled()) {
```
julianladisch, line 215: fullLogger

okapi-core/src/main/resources/log4j2.properties
```diff
@@ -1,19 +1,25 @@
...
+appender.omitted.layout.pattern = %d{HH:mm:ss} [] [] [] [] %-5p %-20.20C{1} %m%n
 
 rootLogger.level = info
 rootLogger.appenderRefs = info
-rootLogger.appenderRef.stdout.ref = STDOUT
+rootLogger.appenderRef.stdout.ref = OMITTED
+
+logger.full = info, FULL
+logger.full.name = org.folio.okapi.util.ProxyContext
+logger.full.additivity = false
```
julianladisch, line 25: Having all log levels been set to info is fine.

But I wonder what a sysop needs to change in this log4j2.properties file to reduce the root level to warn but keep the full (ProxyContext) log level at info.

okapi-core/src/main/java/org/folio/okapi/util/ProxyContext.java
adamdickmeiss: Right. The modules are separated by space. Would it be better to use , ?

okapi-core/src/main/resources/log4j2.properties
steveellis: I guess I would ask why the pattern layout should be consistent between okapi and the modules. I do agree that it breaks the symmetry with the modules but would there be any practical downside in terms of how the logs are used?

I feel like it's an oversight that we're not putting a date in there.

I'll create a JIRA as you suggest and post something on sys ops in slack to get a conversation going on it. In any case I don't think merging what Adam has should be held up by this.

okapi-core/src/main/resources/log4j2.properties
steveellis: Maybe `empty` since it's the pattern for when there's no values inside the brackets. It took me a while to grok "omitted" as well.

steveellis (approved):

okapi-core/src/main/resources/log4j2.properties
adamdickmeiss: Create a JIRA please.

okapi-core/src/main/resources/log4j2.properties
adamdickmeiss: OK. I also slightly prefer "empty" over "basic":

okapi-core/src/main/resources/log4j2.properties
adamdickmeiss: I don't know how to do that. Ideas?

adamdickmeiss: Using 

    rootLogger.level = warn

works for me. Only the REQ/RES lines logged.

julianladisch: I tried `mvn test -Dtest=ProxyTest#testHealth` and it doesn't work but this is not relevant and might be caused by some logging set up in the tests.

When running `java -jar target/okapi-core-fat.jar dev` it works as expected, therefore there's no issue.

okapi-core/src/main/resources/log4j2.properties
```diff
@@ -1,19 +1,27 @@
 status = error
 name = PropertiesConfig
-packages = org.folio.okapi.common.logging
```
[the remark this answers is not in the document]
adamdickmeiss: ok

steveellis (approved):

julianladisch (approved):
Looks good, thanks!

PR #1258, opened 2022-08-30
Use https://folio-registry.dev.folio.org

adamdickmeiss:


dcrossleyau (approved):
I did local beyondgrep.

There is still mention in the doc-zh/guide.md but i presume that we would leave that for translators to generally sync.

PR #1261, opened 2022-09-19
Restrict module names OKAPI-1082

adamdickmeiss:
https://issues.folio.org/browse/OKAPI-1082

At the moment if this is merged, Okapi can not pull from the registry anymore as it has a module that is rejected:

mod-data-import-converter-storage


okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
```diff
@@ -1,34 +1,101 @@
...
+        }
+        if (!isAsciiLetter(c)) {
+          throw new IllegalArgumentException(MODULE_QUOTE + id
+              + "' has non-allowed character at offset " + i);
+        }
+      } else if (!isAsciiLetter(c) && !isAsciiNumeric(c)) {
+        throw new IllegalArgumentException(MODULE_QUOTE + id
+            + "' has non-allowed character at offset " + i);
       }
     }
```
julianladisch, line 79: One if moved out of the loop, less nesting:
```
    if (s.charAt(s.length() - 1) == '-') {
      throw new IllegalArgumentException(MODULE_QUOTE + id
          + "' has hyphen-minus as last character");
    }
    final boolean ui = s.startsWith("folio_");
    int i = ui ? 6 : 1;
    for (; i < s.length(); i++) {
      char c = s.charAt(i);
      if (isAsciiLetter(c) || isAsciiNumeric(c)) {
        continue;
      }
      if (c != '-') {
        throw new IllegalArgumentException(MODULE_QUOTE + id
            + "' has non-allowed character at position " + (i + 1));
      }
      char next = s.charAt(i + 1);
      if (isAsciiNumeric(next)) {
        product = s.substring(0, i);
        productCheck(ui);
        semVer = new SemVer(s.substring(i + 1));
        return;
      }
      if (next == '-') {
        throw new IllegalArgumentException(MODULE_QUOTE + id
            + "' has two hyphen-minuses in sequence at position " + (i + 1));
      }
    }
```

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
```diff
@@ -1,34 +1,101 @@
...
+        c = s.charAt(++i);
+        if (isAsciiNumeric(c)) {
+          product = s.substring(0, i - 1);
+          productCheck(ui);
+          semVer = new SemVer(s.substring(i));
+          return;
+        }
+        if (!isAsciiLetter(c)) {
+          throw new IllegalArgumentException(MODULE_QUOTE + id
+              + "' has non-allowed character at offset " + i);
```
julianladisch, line 73: 1 based position is more readable than 0 based offset.
`at position " + (i + 1)`

okapi-common/src/main/java/org/folio/okapi/common/ModuleId.java
```diff
@@ -1,34 +1,101 @@
 package org.folio.okapi.common;
 
 import java.util.Collection;
+import java.util.List;
 
 public class ModuleId implements Comparable<ModuleId> {
 
+  private static final String MODULE_QUOTE = "ModuleID '";
```
julianladisch, line 8: This is odd. How about `IllegalModuleIdException(String msg)` and `IllegalModuleIdException(String moduleId, String msg)`?

okapi-common/src/test/java/org/folio/okapi/common/ModuleIdTest.java
```diff
@@ -71,6 +69,82 @@ public void test() {
...
+  public void testHyphenNoVersion() {
+    assertEquals("a-x", new ModuleId("a-x").getProduct());
+  }
+
+  @Test
+  public void testDoubleHypen() {
+    assertEquals("ModuleID 'a--x' has non-allowed character at offset 2",
+        assertThrows(IllegalArgumentException.class, () -> new ModuleId("a--x")).getMessage());
+  }
+
```
julianladisch, line 88: A test for the last allowed letter `z` is missing.

PR #1262, opened 2022-09-22
Fix okapi cannot start restricted tenant OKAPI-1121

adamdickmeiss:


okapi-core/src/main/java/org/folio/okapi/managers/TenantManager.java
```diff
@@ -123,6 +126,14 @@ public void setProxyService(ProxyService px) {
...
   public Future<String> insert(Tenant t) {
     String id = t.getId();
+    if (id == null) {
+      return Future.failedFuture(
+          new OkapiError(ErrorType.USER, messages.getMessage("10411")));
+    }
+    if (!TENANT_PATTERN.matcher(id).find()) {
+      return Future.failedFuture(
+          new OkapiError(ErrorType.USER, messages.getMessage("10412", id, TENANT_PATTERN_STRING)));
+    }
```
julianladisch, line 136: The tenant id `pg` should be rejected here.

Then sysOps will get a good error message ("pg is an invalid tenant id because it is reserved by PostgreSQL"), this is to comply with accessibility laws ([WCAG 3.3.1 and 3.3.3](https://www.w3.org/WAI/WCAG21/quickref/#input-assistance)).

Otherwise sysOps will report this as an Okapi bug.

As soon as there are (nontrivial) module implementations that can handle `pg` we may remove this restriction.

okapi-core/src/main/raml/TenantDescriptor.json
```diff
@@ -18,5 +18,6 @@
       "description": "Tenant description",
       "type": "string"
     }
-  }
+  },
+  "required" : [ "id" ]
```
julianladisch, line 22: The "desciption" of "id" should mention "^[a-z][a-z0-9]{0,30}$" and that it cannot be `pg`

PR #1269, opened 2022-10-24
Release 4.14.6 OKAPI-1128

adamdickmeiss:


adamdickmeiss:
3 out of 4 code smells are fixed in master where code is removed (deprecated API as of okapi 5).

PR #1276, opened 2022-11-08
Refer to correct service in login discussion

MikeTaylor:
@adamdickmeiss I think I stopped and fixed a thinko, but please check that I've not misunderstood, and introduced one instead!

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -571,7 +571,7 @@ and they may take different parameters (username and password are the
 most likely, but we can have anything from simple IP authentication to
 complex interactions with LDAP, OAuth, or other systems).
 
-The authorization service returns a token to the client, and the
+The login service returns a token to the client, and the
 client passes this token in a special header in all requests it makes
 to Okapi. Okapi in turn passes it to the authorization module,
 together with information of what modules will be called to satisfy


adamdickmeiss (commented):
There's no mod-login in Okapi guide (no example module as part of Okapi). It's an auth module which does both authorization and authentication.

MikeTaylor:
Then what _does_ mod-login do, if not give you a token in exchange for your credentials?

Or are you saying that the scenario outlined in the Guide is different from the one implemented in FOLIO?

adamdickmeiss:
> Then what _does_ mod-login do, if not give you a token in exchange for your credentials?
> 
> Or are you saying that the scenario outlined in the Guide is different from the one implemented in FOLIO?

mod-login is responsible for that. But mod-login is *not* part of Okapi. Okapi does not mention the FOLIO system. All examples are based on 3 test modules that's part of the distribution.

MikeTaylor:
Ah, I see: the example are not FOLIO examples.

I guess that's OK. But I bet I won't be the last person to be confused by this.

adamdickmeiss:
> Ah, I see: the example are not FOLIO examples.
> 
> I guess that's OK. But I bet I won't be the last person to be confused by this.

The MD of that example module says it all: https://github.com/folio-org/okapi/blob/master/okapi-test-auth-module/descriptors/ModuleDescriptor-template.json

MikeTaylor:
OK, we'll leave this as it is.

PR #1287, opened 2022-12-19
OKAPI-1142: HTTP idle timeout

julianladisch:
Option to configure an HTTP idle timeout for Okapi.

Netty and Vert.x disable it by default.

If enabled Okapi can log the API that is still open after a long time. This helps to investigate https://issues.folio.org/browse/OKAPI-1141

Approach:

HttpClientOptions.setIdleTimeout and configuration option in conf file and as command line parameter.

### DIFF
--- a/dist/okapi.conf
+++ b/dist/okapi.conf
@@ -96,6 +96,10 @@ log4j_config="/etc/folio/okapi/log4j2.properties"
 # This value, if set, overrides the `waitIterations` in the launch descriptor.
 # deploy_waitIterations="60"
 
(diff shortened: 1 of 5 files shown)


adamdickmeiss (commented):
I think this is a very bad idea. If this is enabled by default it will stop "good" things that takes long.

For example reporting.. Which in some cases in our system takes an hour for a response.

Okapi has logWaitMs which does *not* close but warns that a request is ongoing and that Okapi is  still waiting for it.

PR #1303, opened 2023-01-22
[fix] try to close NetClient instance

MiROK-Replication:
It seems that the NetClient instance in `ProcessModuleHandle:waitPortToClose` is not closed. I try to fix this, please review the PR, thanks!


adamdickmeiss:
You are right. We have made a patch here #1304, which should guarantee that it's closed in all cases.

There's no need for socket close anymore as closing the NetClient will also close the net socket.

PR #1305, opened 2023-01-26
OKAPI-1157: Reuse NetClient when checking ports

julianladisch:
ProcessModuleHandle#waitPortOpen and ProcessModuleHandle#waitPortToClose don't close the NetClient.

https://github.com/MiROK-Replication has reported this resource leak on https://github.com/folio-org/okapi/pull/1303 , thank you!

The two methods contain almost the same code so we extract it, and add unit tests.

We don't need a new NetClient for each port test, Vertx allows to reuse a NetClient forever to save resources.

### DIFF
--- a/okapi-core/src/test/java/org/folio/okapi/util/PortCheckerTest.java
+++ b/okapi-core/src/test/java/org/folio/okapi/util/PortCheckerTest.java
@@ -7,11 +7,17 @@ import io.vertx.junit5.Checkpoint;
 import io.vertx.junit5.VertxExtension;
 import io.vertx.junit5.VertxTestContext;
 import java.util.concurrent.atomic.AtomicReference;
+import org.folio.okapi.testing.UtilityClassTester;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 @ExtendWith(VertxExtension.class)
 class PortCheckerTest {
+
+  @Test
+  void utilityClass() {
+    UtilityClassTester.assertUtilityClass(PortChecker.class);
+  }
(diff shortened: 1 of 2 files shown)


okapi-core/src/test/java/org/folio/okapi/util/PortCheckerTest.java
```diff
@@ -0,0 +1,67 @@
...
+    getClosingNetServer(vertx).compose(netServer1 ->
+    getClosingNetServer(vertx).compose(netServer2 ->
+    getClosingNetServer(vertx).compose(netServer3 ->
+    getClosingNetServer(vertx).compose(netServer4 -> {
+      PortChecker.waitPortToClose(vertx, netServer1.actualPort(), 2).onSuccess(x -> c1.flag());
+      PortChecker.waitPortToClose(vertx, netServer2.actualPort(), 2).onSuccess(x -> c4.flag());
+      PortChecker.waitPortToClose(vertx, netServer3.actualPort(), 2).onSuccess(x -> c3.flag());
+      PortChecker.waitPortToClose(vertx, netServer4.actualPort(), 2).onSuccess(x -> c2.flag());
+      return Future.succeededFuture();
+    }))));
```
adamdickmeiss, line 64: missing onComplete with check?

adamdickmeiss (approved):

okapi-core/src/test/java/org/folio/okapi/util/PortCheckerTest.java
adamdickmeiss: thx for explaining

PR #1324, opened 2023-05-11
OKAPI-1169: Sporadic MapExpiryTest failure

julianladisch:
https://issues.folio.org/browse/OKAPI-1169

### DIFF
--- a/okapi-core/src/test/resources/kube-config-bad-syntax.yaml
+++ b/okapi-core/src/test/resources/kube-config-bad-syntax.yaml
@@ -3,7 +3,7 @@ kind: Config
 clusters:
 - name: "folio-eks-2-us-west-2"
   cluster:
-    server: "https://rancher.dev.folio.org/k8s/clusters/c-479xv"
+    server: "https://rancher.ci.folio.org/k8s/clusters/c-479xv"
 
 users:
 - name: "folio-eks-2-us-west-2"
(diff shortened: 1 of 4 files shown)


okapi-common/src/test/java/org/folio/okapi/common/refreshtoken/tokencache/MapExpiryTest.java
```diff
@@ -43,10 +42,13 @@ public void testExpiryPut() {
...
-    tk.put("user2", "v2", System.currentTimeMillis() + 5);
-    Awaitility.await().pollInterval(1, TimeUnit.MILLISECONDS).atMost(10, TimeUnit.MILLISECONDS)
-        .until(() -> tk.get("user2") == null);
+    var expire = System.currentTimeMillis() + 5;
+    tk.put("user1", "v1", expire);
+    tk.put("user2", "v2", expire);
+    while (System.currentTimeMillis() <= expire) {
+      // wait
+    }
+    assertThat(tk.get("user2"), is(nullValue()));
```
steveellis, line 51: Why is the approach with awaitlity brittle? Is it that the time unit of 1 millisecond is too small to work reliably? If so why not just increase the time unit to something a bit more reasonable and keep awaitility instead of the while loop?

steveellis (approved):

PR #1327, opened 2023-09-13
OKAPI-1172: Reduce excessive logging on deploy

julianladisch:
When Okapi deploys a module it logs the complete module descriptor.

This is too verbose. It is not needed because the information is available with `docker inspect`.


adoerrer (approved):

craigmcnally:
I think there might be another place where similar information is logged.  In the OKAPI log I'm looking at I see an INFO message prefixed with: "addAndDeploy"

craigmcnally (approved):

PR #1332, opened 2023-09-26
OKAPI-1175: Upgrade dependencies for Poppy: Vert.x 4.4.5, ...

julianladisch:
Upgrade all dependencies to the latest stable release.

In PostgresQueryTest.java the Fake classes are replaced with Mockito generated mocks. This avoids adding methods to FakeQuery, FakeSqlConnection and FakeHandle whenever Vert.x adds new methods to the parent class.
This makes the test code more easy to understand because they become self-contained.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,7 +58,7 @@
       <dependency>
         <groupId>com.hazelcast</groupId>
         <artifactId>hazelcast</artifactId>
-        <version>4.2.8</version>  <!-- https://github.com/vert-x3/vertx-hazelcast/blob/4.4/pom.xml#L19 -->
+        <version>5.3.2</version>  <!-- https://github.com/vert-x3/vertx-hazelcast/blob/4.x/.github/workflows/ci-4.x.yml -->
       </dependency>
       <!-- END: versions that depend on the vertx-stack-depchain version -->
 


pom.xml
```diff
@@ -43,29 +51,17 @@
...
         <groupId>com.hazelcast</groupId>
-        <artifactId>hazelcast-aws</artifactId>
-        <version>3.4</version>  <!-- https://github.com/hazelcast/hazelcast-aws#requirements -->
-      </dependency>
-      <dependency>
-        <groupId>com.hazelcast</groupId>
-        <artifactId>hazelcast-kubernetes</artifactId>
-        <version>2.2.3</version>  <!-- https://github.com/hazelcast/hazelcast-kubernetes#requirements-and-recommendations -->
+        <artifactId>hazelcast</artifactId>
+        <version>4.2.8</version>  <!-- https://github.com/vert-x3/vertx-hazelcast/blob/4.4/pom.xml#L19 -->
```
steveellis: Can we go to hazlecast 5 @julianladisch? I'm looking at @hjiebsco's PR here: https://github.com/folio-org/okapi/pull/1323. From the looks of this it vertx-hazelcast has been upgraded to handle hazelcast 5 here: https://github.com/vert-x3/vertx-hazelcast/pull/184.

steveellis: Looks like you were making the change just as I was typing the comment.

steveellis (approved):
Upgrading the one class to use Mockito seems to be a little outside of this ticket but clearly using it has improved this class so seems fine.

PR #1346, opened 2024-02-20
OKAPI-1184: Quesnelia dependencies: Vert.x 4.5.3, Micrometer 1.12.2, …

julianladisch:
Upgrade Vert.x from 4.4.8 to 4.5.3.

The Vert.x upgrade causes several deprecations that require small code changes.

The Vert.x upgrade indirectly upgrades Jackson so that we need a new pattern in JsonDecoder.

Upgrade Micrometer from 1.11.5 to 1.12.2 (https://github.com/vert-x3/vertx-micrometer-metrics/blob/c42aed98ef3b92f83eea55463207edd5e4505b2e/pom.xml#L41)

Upgrade log4j from 2.20.0 to 2.22.1.

Keep Hazelcast at 5.3.5.

Remove netty-tcnative-boringssl-static and the OpenSSL configuration. We can use JdkSSL because OpenJDK comes with TLSv1.3.

### DIFF
--- a/okapi-core/src/main/java/org/folio/okapi/MainDeploy.java
+++ b/okapi-core/src/main/java/org/folio/okapi/MainDeploy.java
@@ -12,7 +12,6 @@ import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Verticle;
 import io.vertx.core.Vertx;
-import io.vertx.core.VertxBuilder;
 import io.vertx.core.VertxOptions;
 import io.vertx.core.eventbus.EventBusOptions;
 import io.vertx.core.file.FileSystemOptions;


okapi-core/pom.xml
```diff
@@ -66,11 +71,6 @@
       <groupId>io.vertx</groupId>
       <artifactId>vertx-pg-client</artifactId>
     </dependency>
-    <dependency>  <!-- needed because OpenJDK 8 doesn't support TLSv1.3 -->
```
jakub-id, line 69: @julianladisch is the removal of boringssl related to the vert.x upgrade? If not, maybe it would be better to do it in a separate PR?

okapi-core/src/main/java/org/folio/okapi/util/CorsHelper.java
```diff
@@ -56,7 +56,8 @@ public static void addCorsHandler(Router router, final TenantManager tenantManag
       if (ctx.data().containsKey(DELEGATE_CORS)) {
         ctx.next();
       } else {
-        CorsHandler.create("*")
+        CorsHandler.create()
+            .addRelativeOrigin(".*")
```
jakub-id, line 60: @julianladisch why the pattern has been changed?

jakub-id (approved):
@julianladisch I took this for a spin in my environment and it looks good.

PR #1347, opened 2024-02-21
OKAPI-1081: Reject invalid tenant ids

julianladisch:
https://folio-org.atlassian.net/browse/OKAPI-1081

Implement
https://folio-org.atlassian.net/wiki/display/TC/ADR-000002+-+Tenant+Id+and+Module+Name+Restrictions
so that creating a new tenant via POST /_/proxy/tenants API is rejected unless the tenant id matches
`^[a-z][a-z0-9]{0,30}$`

Legacy tenant ids still work when used in any other API.

### DIFF
--- a/okapi-common/src/test/java/org/folio/okapi/common/GenericCompositeFutureTest.java
+++ b/okapi-common/src/test/java/org/folio/okapi/common/GenericCompositeFutureTest.java
@@ -1,14 +1,11 @@
 package org.folio.okapi.common;
 
-import io.vertx.core.AsyncResult;
 import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
-import io.vertx.core.Handler;
 import io.vertx.core.Promise;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-import java.util.function.Function;
 import org.assertj.core.api.WithAssertions;
 import org.junit.Test;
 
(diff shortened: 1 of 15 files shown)


jakub-id:
@julianladisch while this avoids problems with existing tenant data. it is still a breaking change of the API. If we need to bump Okapi version to 6.0.0, I'd like us to release 5.3.0 first. Platform module release deadline for Quesnelia is March 1.

hjiebsco (commented):
https://folio-org.atlassian.net/wiki/spaces/DD/pages/1779867/Tenant+Id+and+Module+Name+Restrictions mentions `Okapi and the modules should provide APIs and/or scripts to do the migration.` Is that ready? Can we provide a configuration to allow user to continue use the old `^[a-z0-9_-]+$` validation to avoid this breaking change?

hjiebsco:
> Legacy tenant IDs can still be used. The restriction only affects new tenant IDs created via the API. There's no need to migrate legacy tenant IDs. Therefore this breaking change doesn't affect existing tenant IDs. The breaking change only affects tenant IDs that people will create after the upgrade.

From users' perspective, for whatever reason, they might need or prefer to name new tenant id consistently with old tenants, or they might need to recreate old tenants with the same ids. It would be better to provide configuration to allow these.

hjiebsco:
> > Legacy tenant IDs can still be used. The restriction only affects new tenant IDs created via the API. There's no need to migrate legacy tenant IDs. Therefore this breaking change doesn't affect existing tenant IDs. The breaking change only affects tenant IDs that people will create after the upgrade.
> 
> From users' perspective, for whatever reason, they might need or prefer to name new tenant id consistently with old tenants, or they might need to recreate old tenants with the same ids. It would be better to provide configuration to allow these.

FYI that was just a suggestion. No objection if you want to merge as is.

okapi-core/src/main/java/org/folio/okapi/managers/InternalModule.java
```diff
@@ -750,22 +751,22 @@ private String location(ProxyContext pc, String id, String baseUri, String s) {
...
+          .compose(x -> tenantManager.insert(new Tenant(td)))
+          .map(x -> location(pc, td.getId(), null, Json.encodePrettily(td)));
     } catch (DecodeException ex) {
       return Future.failedFuture(new OkapiError(ErrorType.USER, ex.getMessage()));
     }
   }
 
+  static Future<Void> validateTenantId(TenantDescriptor td) {
+    if (td.getId() == null || td.getId().isEmpty()) {
+      td.setId(("t" + UUID.randomUUID().toString().replace("-", ""))
```
steveellis, line 764: Maybe break creation of the uuid tenant id out into its own method called `tryCreateTenantId` or simlar and do validation separately since they seem like different operations to me.

steveellis (approved):

PR #1352, opened 2024-05-16
OKAPI-1186: Add ChattyResponsePredicate to okapi-common

julianladisch:
See
* https://github.com/folio-org/mod-user-import/pull/109/files
* https://folio-org.atlassian.net/browse/MODUIMP-93
* https://vertx.io/docs/vertx-web-client/java/#_creating_custom_failures

ChattyResponsePredicate is the same as Vert.x' ResponsePredicate but with response body included in the Exception message.

The body explains why the request failed. The HTTP status (400, 500) is not sufficient to investigate the cause.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,10 @@
+## 6.0.2 2024-06-06
+
+Quesnelia version.
+
+Fix:
+* [OKAPI-1187](https://folio-org.atlassian.net/browse/OKAPI-1187) Illegal base64 character 5f on token with umlaut
+
 ## 6.0.1 2024-04-05
 
 Quesnelia version.
--- a/okapi-common/src/main/java/org/folio/okapi/common/OkapiToken.java
+++ b/okapi-common/src/main/java/org/folio/okapi/common/OkapiToken.java
@@ -1,5 +1,7 @@
 package org.folio.okapi.common;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 import io.vertx.core.json.DecodeException;
 import io.vertx.core.json.JsonObject;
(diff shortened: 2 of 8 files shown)


okapi-common/src/main/java/org/folio/okapi/common/ChattyResponsePredicate.java
```diff
@@ -0,0 +1,433 @@
...
+   * the concatenation result as message.
+   *
+   * @see ResponsePredicate#create(java.util.Function, ErrorConverter)
+   */
+  public static final ErrorConverter CONVERTER = ErrorConverter.createFullBody(result -> {
+    var body = result.response().bodyAsString();
+    if (body != null && body.length() > MAX_LENGTH) {
+      body = body.substring(0, MAX_LENGTH / 2) + "…"
+          + body.substring(body.length() - (MAX_LENGTH - 1) / 2);
+    }
```
steveellis, line 29: "Removing characters from the middle" (from the comment above)... Can you explain why you want to "remove characters from the middle". Is it because there is an assumption that the middle will contain extraneous information, but not the beginning or end?

steveellis: I see the purpose but I wonder if it should be the default behavior to truncate this middle part. It seems like a premature optimization perhaps and a bit arbitrary to base it off of character count. Could it be perhaps made a boolean option of the class such that users of `ChattyResponsePredicate` decide whether to reduce the body size if they know their error body is going to be large?

But this is just an idea for you to consider. I don't see this as a major problem .

steveellis (approved):

PR #1359, opened 2024-08-30
OKAPI-1191: NumberFormatException in timerId halts Okapi

julianladisch:
https://folio-org.atlassian.net/browse/OKAPI-1191

Avoid NumberFormatException in timerId.

Approach:

Find `_` from the end (not from the beginning) because the tenantId may contain `_`.

Wrap parsing into try-catch to gracefully handle all parse errors.

Move the timerId code into TenantProductSeq class.

Use product_seq only in module descriptor and in APIs; use tenant_product_seq in all other places.

Rename timerId variable to tenantProductSeq or productSeq.

### DIFF
--- a/okapi-core/pom.xml
+++ b/okapi-core/pom.xml
@@ -127,6 +127,15 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <!-- raml-tester uses vulnerable commons-io 2.2:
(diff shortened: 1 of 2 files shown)


steveellis (approved):
Nice hardening of okapi timer ids!

PR #1366, opened 2024-11-29
OKAPI-1196: ChattyHttpResponseExpectation

julianladisch:
https://folio-org.atlassian.net/browse/OKAPI-1196

Vert.x has deprecated ResponsePredicate, we need to use HttpResponseExpectation instead:
https://github.com/vert-x3/vertx-web/blob/4.5.11/vertx-web-client/src/main/java/io/vertx/ext/web/client/predicate/ResponsePredicate.java#L49

Therefore we need to deprecate our ChattyResponsePredicate and provide a replacement:

ChattyHttpResponseExpectation

### DIFF
--- a/okapi-common/src/test/java/org/folio/okapi/common/ChattyHttpResponseExpectationTest.java
+++ b/okapi-common/src/test/java/org/folio/okapi/common/ChattyHttpResponseExpectationTest.java
@@ -135,3 +135,4 @@ class ChattyHttpResponseExpectationTest {
     }));
   }
 }
+


okapi-common/src/test/java/org/folio/okapi/common/ChattyHttpResponseExpectationTest.java
```diff
@@ -0,0 +1,137 @@
...
+    .compose(httpServer -> WebClient.create(vertx)
+        .post(httpServer.actualPort(), "localhost", "/")
+        .send()
+        .expecting(expectation))
+    .onComplete(vtc.failing(e -> {
+      assertThat(e.getMessage(), allOf(containsString(contentType), endsWith(": Body and Soul")));
+      vtc.completeNow();
+    }));
+  }
+}
```
steveellis, line 137: No new line at EOF

steveellis (approved):

PR #1372, opened 2025-02-14
Explain tailing slash in deployment url property

julianladisch:


adamdickmeiss (approved):
Have you tried or tested that trailing slash is passed through if given ?

adamdickmeiss (approved):

PR #1388, opened 2025-05-15
OKAPI-1200 fix ModuleTest.testInitdatabaseBadCredentials fails

adamdickmeiss:
On initdatabase failure we do not need stop the Verticle. It is stopped already because a failure is returned.


adamdickmeiss:
> Ok.
> 
> Calling `vertx.close()` is not the gentle way to stop the program. A gently way might be `promise.handle(new StopVerticle(initMode + " has been completed."))` where StopVerticle extends NoStackTraceException.

Tried that here : #1389  .

I think the whole init stuff should not have been Verticle based.

PR #1393, opened 2025-05-22
OKAPI-1212: okapi-common: Upgrade to Vert.x 5

julianladisch:
https://folio-org.atlassian.net/issues/OKAPI-1212

The shared library okapi-common needs to migrate Vertx from 4 to 5.

It’s required for Trillium: https://folio-org.atlassian.net/wiki/spaces/TC/pages/730891059/Trillium#Trillium-ThirdPartyLibraries/Frameworks

See migration guide: https://folio-org.atlassian.net/wiki/spaces/FOLIJET/pages/1003847853/Migration+to+Vertx+5

FOLIO adopots the Eureka platform: https://folio-org.atlassian.net/wiki/spaces/TC/pages/518946885/0010-eureka
Eureka no longer needs and no longer uses the Okapi server.
Therefore the okapi-core directory (= Okapi server) and all server related content is dropped in this PR.


okapi-common/diff
steveellis, line 1: Do you want this file?

okapi-common/pom.xml.releaseBackup
steveellis, line 1: Do you want this file?

steveellis (approved):
Thanks for taking care of this Julian!

PR #1404, opened 2025-08-08
OKAPI-1214 okapi 6.3.0 for trillium

adamdickmeiss:


adamdickmeiss:
Should dependencies also be updated in this PR?
We used to do it separately, but #1402 suggests not.

jakub-id (approved):

jakub-id:
@julianladisch OK to merge?

PR #1443, opened 2026-04-01
Release 7.0.3

julianladisch:
7.0 is the Trillium release.

We don't need 7.1.0 for Trillium.


adamdickmeiss (commented):
Ok. So 7.1.0 should be unreleased?

adamdickmeiss (approved):

PR #1450, opened 2026-06-10
OKAPI-1248: Use GitHub Actions Workflows for Maven replacing Jenkinsfile

julianladisch:
https://folio-org.atlassian.net/browse/OKAPI-1248


dcrossleyau (approved):
Excellent Julian, thanks for sorting this out.

I tried adding the "docker-health-command" as was done in the old Jenkinsfile, but that build-and-test failed. So removed it.
Maybe you know what is happening there.

I do not know what is wrong with the Sonar thing, but that should not hold up merging this.

The other jobs are okay.

If you need help with the branch protection [configuration](https://github.com/folio-org/.github/blob/master/README-maven.md#install-the-caller-workflow) then i can assist.

julianladisch:
Okapi needs `runArgs = 'dev'` for the docker-health-command.

We test the Docker container in the OkapiIT.java integration test instead.

Thank you for fixing the maven.yml and adding the docker-label-documentation.

dcrossleyau:
Jason fixed the weird Sonar problem. It wanted the project key changed at the Sonar end.

